PR #10, opened 2016-07-26
cmake: Fix symbol detection on non-standard include paths

jktjkt:
We're still in the FindModule scope and the global include directories
therefore very likely do not contain the required path.

I used list(INSERT ... 0) and list(REMOVE_AT ... 0) instead of the more
intuitive APPEND and REMOVE_ITEM because REMOVE_ITEM actually removes
all occurrences. That might affect the global state which is exactly not
what a find_package ought to do. If the minimal required CMake was
2.8.6+, one could have used CMakePushCheckState, but it isn't available
in 2.6.


michalvasko:
Hi,
it seems you are correct in this case, thanks.

Regards,
Michal

PR #25, opened 2016-09-25
fix 2 bugs

qmm161:
fix bug: 1. session not refreshed when sysrepo commit return SR_ERR_VALIDATION_FAILED. 
2. free str would crush system because str pointed to path in line 472 or 496


rkrejci:
Hi,
please keep the bugfixes in separated commits (it can be in a single pull request, but definitely as a separate commits since it fixes different things).

PR #29, opened 2016-09-28
fix bug: free str would crush system when str point to path

qmm161:


qmm161:
is this pull request ok?

rkrejci:
Thanks!

PR #37, opened 2016-10-03
Fix linking of tests

jktjkt:
These libraries might not be available from the global systemwide
include paths, and with CMake we already know where to find them.


rkrejci:
Merged as 8c512219a5

PR #40, opened 2016-10-07
cmake: Allow specifying several libs in sysrepo.pc

jktjkt:
As a side effect of pull request sysrepo/sysrepo#389, sysrepo.pc
pkg-config file is changed so that it now specifies both libsysrepo.so
and libsysrepo-engine.so for linking.  This patch adapts to that change
(if it gets merged) so that netopeer2 now supports linking with sysrepo
again.

I am not completely happy about this. I asked on #cmake IRC channel and
apparently The Modern Way™ would be for sysrepo to install a special
fancy cmake file which one could use to bypass pkg-config altogether. I
asked because I got puzzled by the suggested dance to take pkg-config's
results and call cmake's find_library on top of them.

Another alternative would be to add link_directories(), but that would
require calling find_module(SYSREPO) prior to defining targets -- please
indicate whether you prefer this LIST() dance, or whether you're OK with
moving this order to detect libs at first, and only then define targets.

### DIFF
--- a/CMakeModules/FindSYSREPO.cmake
+++ b/CMakeModules/FindSYSREPO.cmake
@@ -4,23 +4,25 @@
 #  SYSREPO_DEFINITIONS - Compiler switches required for using SYSREPO
 
 find_package(PkgConfig)
-pkg_check_modules(PC_SYSREPO QUIET sysrepo)
-set(SYSREPO_DEFINITIONS ${PC_SYSREPO_CFLAGS_OTHER})
-
-find_path(SYSREPO_INCLUDE_DIR sysrepo.h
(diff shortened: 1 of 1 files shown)


rkrejci:
We actually use the cmake code from sysrepo - our `FindSYSREPO.cmake` is `sysrepo-config.cmake` from its root. I believe that this should be done first in sysrepo to allow all other projects based on sysrepo using the same cmake code approved by the sysrepo developers. That's probably very close to that suggested Modern Way™, but the file is not installed, just available in the sysrepo repository.

I would prefer to not use the `link_directories()`, the better way seems to me to use exactly the same linker options that sysrepo (detected from `.pc` file or somehow else) declares that should be used.

jktjkt:
OK, based on discussion with upstream, this won't be necessary. Closing.

PR #44, opened 2016-10-11
fix cli crash when executing 'copy-config --target startup  --source running'

fanchanghu:
Hi,
I try to fix a bug that cli crash when executing 'copy-config --target startup  --source running'.


michalvasko:
Hi,
thanks, I modified your patch slightly before merging.

Regards,
Michal

PR #47, opened 2016-10-26
fix edit-config & get-config failed for leafref to decimal64

fanchanghu:


fanchanghu:
as the code, only `decimal64` need get some information(fraction-digits) from schema when converting between `sr_val_t*` and `char*`. when type is leafref, you must get fraction-digits from the target of leafref.

fanchanghu:
I review the code, it seems `LY_TYPE_BITS` also need consider leafref and I push a new commit for it.

rkrejci:
Seems still not enough - going through leafrefs, you can get into union type. It seems that having a libyang function to get correct type definition would be fine. I'll try to create something.

rkrejci:
`lyd_leaf_type()` from the current libyang devel (0.11.32) should fit the needs for bits. In case of decimal64, the fraction-digits is not actually needed at all.

If you are fine with that, I will fix it on my own.

fanchanghu:
yes, it's better to fix it in libyang, thanks.

Issue #58, opened 2016-11-07
Error occurred when use get-config to get state data 


zzyyll2:
If there ars some state yang model and data(config=false) in sysrepo,when using get --filter-xpath=/stateModel:stateModel in netopeer2-cli，it get some correct data,but using get-config --source running,it response error msg back.
like this,dataAid is a leaf with config=false property.
> get-config --source running
ly ERROR: Unknown element "dataAid". (/stateModel:stateModel/dataAid)
nc ERROR: Failed to parse <get-config> reply.
cli_send_recv: Failed to receive a reply.

michalvasko:
Hi,
it should work now, please verify.

Regards,
Michal

PR #60, opened 2016-11-08
fix issue #58

zzyyll2:
YANG model like this
module status{
  namespace "urn:ccn:params:xml:ns:yang:hw:status";
  prefix "status";
  
  container status{
    leaf dataAid{
      description "the value of this leaf will plus 1 every time that configuration data is modified.";
      type uint32;
	  config false;
	  default 1;
    }
  }
}

when get --filter-xpath=/status:status,it responsed 1,but get-config --source running ,error occurred.

This code fix the problem.

### DIFF
--- a/server/op_get_config.c
+++ b/server/op_get_config.c
@@ -32,7 +32,8 @@
 
 /* add subtree to root */
 static int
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
there is no need to filter out state data, sysrepo session can do it for us.

Regards,
Michal

PR #122, opened 2017-04-06
cmake: Use absolute path for keystored's key directory

jktjkt:
The correct dir doesn't live underneath /usr on a typical system, so it
should be specified as an absolute path like this.

### DIFF
--- a/keystored/CMakeLists.txt
+++ b/keystored/CMakeLists.txt
@@ -25,7 +25,7 @@ set(KEYSTORED_VERSION 0.1.1)
 
 # config variables
 if (NOT KEYSTORED_KEYS_DIR)
-    set(KEYSTORED_KEYS_DIR "${CMAKE_INSTALL_PREFIX}/etc/keystored/keys")
+    set(KEYSTORED_KEYS_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/keystored/keys")
 endif()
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
the way it is now it uses the same prefix for all the files and binaries, which seems to me is the best approach. It's true that `keystored` creates only this directory during it's installation, the plugin is copied into the correct path based on *sysrepo* plugin directory, so it cannot be customized. Still, if we added, let's say, some documentation into `share`, I think the prefix should be the same, keystored directory by default is `/usr/local/etc/keystored` and documentation should go into `/usr/local/share`, the prefix should be the same. So I do not really understand the problem, can you justify this change better please?

Regards,
Michal

jktjkt:
I've pushed an updated version which relies on ``CMakeInstallDirs.cmake`` that implements correct logic for both you and me. Please check the commit message of b238e23ee0d467ed025a661c71504e85c3d4d02d.

michalvasko:
Hi,
this way it will be fine. Just a side note, shouldn't '/' be used as prefix for production use? Nevertheless, it should work fine for all cases now.

Regards,
Michal

jktjkt:
> Just a side note, shouldn't `/` be used as the prefix for production use?

Linux distributions that I know all invoke cmake with ``-DCMAKE_INSTALL_PREFIX=/usr``.

Thanks for merging this.

PR #128, opened 2017-04-13
fix edit-config using model with ordered-by user attr under leaf-list report error

yanjunxi:
repeat step:
1. Install the module of module.yang. It will success.
2. Use edit-config in netopeer2-cli using init.xml to config data. It will report OK.
3. Use edit-config in netopeer2-cli using modify.xml to config data. It will report error like this:
    "message:  edit_get_move: memory allocation failed (Unknown error -8191)"

module.yang:
```
module UserByOrderTest{
	namespace "urn:ietf:params:xml:ns:yang:userbyordertest";
	prefix UserByOrderTest;
	
	container root{
		leaf-list ff{
			type string;
			ordered-by user;
		}
	}
}
```

init.xml:
```xml
<root xmlns="urn:ietf:params:xml:ns:yang:userbyordertest"
	xmlns:xc="urn:ietf:params:xml:ns:netconf:base:1.0">
	<ff>aa</ff>
	<ff>bb</ff>
	<ff>cc</ff>
</root>
```

modify.xml
```xml
<root xmlns="urn:ietf:params:xml:ns:yang:userbyordertest"
	xmlns:xc="urn:ietf:params:xml:ns:netconf:base:1.0"
	xmlns:yang="urn:ietf:params:xml:ns:yang:1">
	<ff xc:operation="create" yang:insert="before" yang:value="bb">dd</ff>
	<ff xc:operation="create" yang:insert="after" yang:value="aa">ee</ff>
</root>
```

[issue.txt](https://github.com/CESNET/Netopeer2/files/918402/issue.txt)

### DIFF
--- a/server/op_editconfig.c
+++ b/server/op_editconfig.c
@@ -99,7 +99,7 @@ edit_get_move(struct lyd_node *node, const char *path, sr_move_position_t *pos, 
                     *pos = SR_MOVE_AFTER;
                 }
             } else if (!strcmp(attr_iter->name, name)) {
-                if (asprintf(rel, format, path, attr_iter->value_str)<0) {
+                if (asprintf(rel, format, path, attr_iter->value_str) < 0) {
                     ERR("%s: memory allocation failed (%s) - %s:%d",
                         __func__, strerror(errno), __FILE__, __LINE__);
                     return EXIT_FAILURE;
--- a/server/op_editconfig.c
+++ b/server/op_editconfig.c
@@ -99,7 +99,7 @@ edit_get_move(struct lyd_node *node, const char *path, sr_move_position_t *pos, 
                     *pos = SR_MOVE_AFTER;
                 }
             } else if (!strcmp(attr_iter->name, name)) {
-                if (asprintf(rel, format, path, attr_iter->value)) {
+                if (asprintf(rel, format, path, attr_iter->value_str)<0) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you were right about the error, but I don't know why you changed `value` to `value_str`. But it should work now (I tested your use-case).

Regards,
Michal

yanjunxi:
because struct lyd_attr is changed like the follow, value's type is not char* now:

struct lyd_attr {
    struct lyd_node *parent;         /**< data node where the attribute is placed */
    struct lyd_attr *next;           /**< pointer to the next attribute of the same element */
    struct lys_ext_instance_complex *annotation; /**< pointer to the attribute/annotation's definition */
    const char *name;                /**< attribute name */
    const char *value_str;           /**< string representation of value (for comparison, printing,...), always corresponds to value_type */
    lyd_val value;                   /**< node's value representation, always corresponds to schema->type.base */
    uint16_t value_type;             /**< type of the value in the node, mainly for union to avoid repeating of type detection,
                                          if (schema->type.base == LY_TYPE_LEAFREF), then value_type may be
                                          (LY_TYPE_LEAFREF_UNRES | leafref target value_type) and (value.leafref == NULL) */
};

michalvasko:
Hi,
oh, you are correct, but please stick to our code formatting (add spaces around `<`). I can merge it then.

Regards,
Michal

PR #129, opened 2017-04-15
server BUGFIX failed to handle notification with nested data node

fanchanghu:
Hi,
I found that  notification with nested data node could not be sent by server, and then try to fix it.


michalvasko:
Hi,
thanks, bug reproduced and fix seems to be working.

Regards,
Michal

PR #135, opened 2017-05-02
BUGFIX typo in cmake

mislavn:
Docker build for Arch linux would fail because of this typo, all other dockerfiles would pass.


michalvasko:
Actually, this was the removed sysrepoctl init operation, so it should be removed from here as well. Does it work now?

Regards,
Michal

mislavn:
> Actually, this was the removed sysrepoctl init operation, so it should be removed from here as well. Does it work now?

Yes. I will close the pull request.

Issue #141, opened 2017-05-17
a root CMakeLists.txt should be created to allow building both cli & server at same time or not


ykoehler:
When a package has sub-package, it should provide a CMakeLists.txt to allow building all its parts or only some sub-parts.  Here, we have a CMakeLists.txt in ./server and another one in ./cli and we need to invoke the CMake twice to build both. 

More appropriate would be a root CMakeLists.txt files allowing building of server & cli, just server or just cli.

PR #147, opened 2017-05-25
Fix CESNET/Netopeer2#141 by allowing building both cli/server component

ykoehler:


rkrejci:
I'll have to update also documentation (build instructions). How the user is supposed to build only server or only cli?

ykoehler:
I have not put a flag to build only one or the other, I will add that.

PR #151, opened 2017-06-13
server BUGFIX skip default nodes when copy-config

fanchanghu:
Hi,
Netopeer2 will create default nodes when using copy-config, this request is used to fix it.


fanchanghu:
Hi,
it seems that CI failed duo to any other reason but not this request itself, how can I rerun the CI to make it pass?
@michalvasko

PR #168, opened 2017-08-15
server CHANGE support anyxml/anydata in rpc output and notification

fanchanghu:
Hi,
This request is to make the `anyxml`/`anydata` node be created correctly.


michalvasko:
Hi,
seems fine, we have never tried this, thanks.

Regards,
Michal

PR #169, opened 2017-08-16
TLS listen does not work with child enumeration xpath

zyxel-jimmywang:
TLS listen does not work because the enumeration with the xpath "/ietf-netconf-server:netconf-server/listen/endpoint[tls]/*".


michalvasko:
Hi,
it was a typo, thanks.

Regards,
Michal

Issue #170, opened 2017-08-23
Netopeer2 CLI compile issue on glibc


mislavn:
Hi,

I encountered some problems while compiling Netopeer2 CLI on uClibc.
1) mkstemps not supported (this is not a problem, similar to [sysrepo](https://github.com/sysrepo/sysrepo/pull/910/commits/2a6978c79f069218747325cae1e2faced8c52644) ).
2) eaccess not supported.

Can you tell is it safe to replace eaccess with access?

Or do you know some other solution?

I compiled Netopeer2 CLI with `#define eaccess access` and it works but I am not sure how to properly test it.

Regards,
Mislav

michalvasko:
Hi Mislav,
it should be fine using `access` instead `eaccess`, they are likely to give similar result (EUID equals UID) and even if not, it is used mostly for checks, wrong return value should not cause any actual error.

Regards,
Michal

PR #171, opened 2017-08-23
Uclibc

mislavn:
Hi,

this is a fix for the github issue https://github.com/CESNET/Netopeer2/issues/170

@michalvasko  can you take a look.

Regards,
Mislav


michalvasko:
Hi Mislav,
from failing build it seems `fcntl.h` must always be included, not just when `mkstemps` does not exist on the target system.

Regards,
Michal

PR #172, opened 2017-08-25
server BUGFIX replace sr_get_iter with sr_get_items

mislavn:
Hi,

This fixes the github issue https://github.com/sysrepo/sysrepo/issues/920.

So when a fresh Sysrepo/Netopeer2 docker container has been created with the default cmake values, netopeer2 crashes. To fix it I need to disable tls features in ietf-netconf-server and add this patch.

It seems that with sr_get_items_iter sysrepo thinks it has more values when it is finished while with sr_get_items it has allocated all of the values at the start.

@rkrejci I am not sure if I can use sr_get_items instead of sr_get_items_iter because of the function module_change_resolve. Can you take a look.

Regards,
Mislav

### DIFF
--- a/server/ietf_netconf_server.c
+++ b/server/ietf_netconf_server.c
@@ -1161,8 +1161,8 @@ feature_change_ietf_netconf_server(const char *feature_name, bool enabled)
 {
     int rc, rc2 = 0;
     const char *path = NULL;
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
it should be fixed now making your fix unnecessary.

Regards,
Michal

PR #207, opened 2017-12-28
server BUGFIX subtree filter error for two content match nodes

fanchanghu:
Hi,
Operation `get` with subtree filter will get wrong nodes if it has two content match nodes, of which the first matched and second not, this request is used to fix it. I also add a test case for this case.


michalvasko:
Thanks!

Regards,
Michal

PR #209, opened 2017-12-30
double free issue about notification 

ustcpxy:
Hi ,
please have a look.

in function np2srv_ntf_clb end, it will free ntf,

but, in np2srv_ntf_send, if call op_filter_get_tree_from_data failed, it would free ntf also.

in this case, the ntf would be freed twice. it would case crash.


michalvasko:
Hi,
I have merged the relevant commit, thanks.

Regards,
Michal

PR #232, opened 2018-03-19
Listening on both ipv6 and ipv4 addresses

morganzhh:
Using "::" as the binding address could listening on local ipv6 and local ipv4 addresses.

Reference

https://github.com/nodejs/node/issues/9390

List here

OS | Listen IPv6 address | Does equivalent IPv4 auto listened
-- | -- | --
Windows 7Pro/10 | :: | YES
Mac OS X EI Capitan 10.11.6 | :: | YES
Ubuntu 14.04/16.04 | :: | YES
Windows 7Pro/10 | ::1 | NO
Mac OS X EI Capitan 10.11.6 | ::1 | NO
Ubuntu 14.04/16.04 | ::1 | NO
Windows 7Pro/10 | ActualIPv6Address | NO
Mac OS X EI Capitan 10.11.6 | ActualIPv6Address | NO
Ubuntu 16.10 | ActualIPv6Address | NO

### DIFF
--- a/server/main.c
+++ b/server/main.c
@@ -1029,7 +1029,7 @@ server_init(void)
         if (nc_server_add_endpt("main", NC_TI_LIBSSH)) {
             goto error;
         }
-        if (nc_server_endpt_set_address("main", "0.0.0.0")) {
+        if (nc_server_endpt_set_address("main", "::")) {
             goto error;
         }
         if (nc_server_endpt_set_port("main", 830)) {


michalvasko:
Hi,
I actually seem to remember that we initially listened on `::` but there were problems with it if you compiled it an run on a machine without IPv6 support. This could be detected by CMake but I am not sure it is worth it. If you do not want to listen on `0.0.0.0`, just enable server configuration and you can listen on whatever interfaces you want, that is how it was intended.

Regards,
Michal

morganzhh:
Hi Michal,

I forgot the "ipv6 not enable part", then the configuration file is the right way to do.

Thanks,
Morgan

PR #234, opened 2018-03-28
cli BUGFIX segfault when get-schema returns empty data

leonshaw:
When the target of get-schema is not found, some device is replying some thing like:
```
<?xml version="1.0" encoding="UTF-8"?>
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="100">
    <data/>
</rpc-reply>
```
causing segment fault in netopeer2-cli. Add checking to fix this.


michalvasko:
Hi,
no server should respond with such a message but we should be able to handle it. Thanks.

Regards,
Michal

PR #235, opened 2018-04-10
Refresh the session to prevent sysrepo returning cached data for the server certificate

darshana-jk:
This issue was observed when the server certificate is overwritten by a new one with an edit-config.
Sysrepo updated the ietf-keystore.running correctly, but still sent the old certificate when SSL accept.
Session refresh solved the issue. Please review and add.


rkrejci:
@michalvasko - maybe this issue is present at several other places (before calling `np2srv_sr_get_item*()` functions), can you please check it?

darshana-jk:
Updated the code.

rkrejci:
merged manually

PR #248, opened 2018-05-19
Fix enabling a feature that depends on a feature in a subsequent schema

alangefe:
### Description
`np2srv_init_schemas()` walks through schemas in the order they are returned from sysrepo via `np2srv_sr_list_schemas()` and attemps to enable features that sysrepo indicates are supposed to be enabled.

However, if a feature in schema `n` depends on a feature that is found in schema `n + 1` (or any subsequent schema, for that matter), enabling the feature in schema `n` will fail because the feature in schema `n + 1` has not yet been enabled.

The additional unit test included in this commit demonstrates the problem.

### Solution
This commit implements a solution as follows:
1. When walking through the schemas returned from sysrepo, keep a list containing all the features that will need to be enabled.
2. Now walk through all the schemas attempting to enable the recorded features. Some features may fail to enable due to depending on other features that have not yet been enabled. So continue walking through the schemas attempting to enable features as long as (1) we are able to make progress and enable at least one new feature, or (2) there are no more features to enable.

This approach should allow us to enable all of the required features regardless of how far apart they are on a dependency graph.

### DIFF
--- a/server/main.c
+++ b/server/main.c
@@ -754,10 +754,19 @@ np2srv_init_schemas(void)
     char *data = NULL;
     const struct lys_module *mod;
     sr_schema_t *schemas = NULL;
-    size_t count, i, j;
+    size_t schema_count, i, j;
+
+    size_t module_feature_count = 0;
+    struct mod_feat {
+        const struct lys_module *mod;
+        const char *enabled_feature;
+    };
+    struct mod_feat *mod_feat_array;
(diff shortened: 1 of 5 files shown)


michalvasko:
Hi,
thanks, generally nice and working, I will just tune the output messages so that no redundant ones are shown.

Regards,
Michal

PR #251, opened 2018-05-22
cmake: Do not clutter up the source dir during build

jktjkt:
As usual, all build artefacts should go outside of the source tree.


michalvasko:
No problem, thanks anyway.

Regards,
Michal

PR #254, opened 2018-05-23
ietf_system: do not error on unsupported fields

rjarry:
When modifying ietf-system/authentication/user items on other fields than authorized-key, there is an error:

```
 Path "/ietf-system:system/authentication/user[name='root']/password" modified.
 Internal error (.../server/ietf_system.c:79)
```

Since netopeer does not care about these fields, it should ignore them.


rjarry:
Thanks!

PR #255, opened 2018-05-23
Improve experience when cross-compiling (permissions, key management)

jktjkt:
We're using Buildroot for generating images which run on many devices. Buildroot performs the build as a regular user, and therefore one cannot call `sysrepoctl` with options for specifying a particular owner of a module. The build also has no access to the target system's SSH keys (they are generated during the first boot).

This patch series adds two options which make this situation doable. With these in place, it's reasonably straightforward to use Buildroot's native facilities for enforcing correct permissions as well as using systemd's units for proper key management. Yay.

I'll send patches which add the entire NETCONF server stack to Buildroot once these changes hit a release.

### DIFF
--- a/keystored/CMakeLists.txt
+++ b/keystored/CMakeLists.txt
@@ -102,6 +102,12 @@ if (NOT CHMOD_EXECUTABLE)
     message(FATAL_ERROR "Unable to find chmod, set CHMOD_EXECUTABLE manually.")
 endif()
 
+# Command line options to be passed to `sysrepoctl` when working with modules
+# which should only be accessible by an administrator
+if (NOT SYSREPOCTL_ROOT_PERMS)
+    set(SYSREPOCTL_ROOT_PERMS "-o root:root -p 600")
+endif()
+
 # create the keys directory with correct permissions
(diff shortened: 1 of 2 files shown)


jktjkt:
FYI, here's a simple systemd unit which implements key copying:

```systemd
[Unit]
Description=Prepare SSH keys for the NETCONF server
Before=sshd.service
Before=sysrepo-plugind.service
ConditionPathExists=!/etc/keystored/keys/ssh_host_rsa_key.pub.pem

[Service]
Type=oneshot
ExecStartPre=/usr/bin/ssh-keygen -A
ExecStart=/bin/cp /etc/ssh/ssh_host_rsa_key /etc/keystored/keys/ssh_host_rsa_key.pem
ExecStart=/bin/openssl rsa -pubout -in /etc/keystored/keys/ssh_host_rsa_key.pem -out /etc/keystored/keys/ssh_host_rsa_key.pub.pem

[Install]
WantedBy=multi-user.target
```

It works.

michalvasko:
Hi,
no problem in merging these changes if they are useful for your.

Regards,
Michal

PR #257, opened 2018-05-24
make auth keys settings accessible for call home

frimpler:
Hi,
I made some small add-ons in order to get "call home" using netopeer2-cli working with public key authentication. These changes were necessary because the keys added by "auth keys add ..." command were not used for the "call home" ssh connection;  only password / interactive authentication was possible.

Please have a look at it. 

Regards,
Frank


michalvasko:
Hi Frank,
seems fine, no one will likely mind that the keys will always be used for both normal connections and Call Home ones, thanks.

Regards,
Michal

PR #258, opened 2018-05-24
Keystored cross compilation fixes

samuel-gauthier:
Using those patches, it becomes possible to compile keystored without installing anything on the compilation host.

- ENABLE_CONFIGURATION is the same as what is already available for the server
- KEYSTORED_KEYS_DIR for the server can be overridden from outside, even if it is not detected by cmake

Once the compilation is done, everything has to be configured manually before starting netopeer / sysrepo where it was installed. It looks like this.

Server configuration:
> sysrepoctl -i --yang /path/to/ietf-ssh-server.yang
> sysrepoctl -i --yang /path/to/ietf-tls-server.yang
> 
> sysrepoctl -i --yang /path/to/ietf-netconf-server.yang
> for feature in listen ssh-listen tls-listen call-home ssh-call-home tls-call-home; do
>        sysrepoctl -m ietf-netconf-server -e $feature
> done
> 
> sysrepoctl -i --yang /path/to/ietf-system.yang 
> for feature in authentication local-users; do
>        sysrepoctl -m ietf-system -e $feature
> done
> sysrepocfg -d startup -i /path/to/stock_config.xml ietf-netconf-server

Keystored configuration:
> sysrepoctl -i --yang /path/to/ietf-keystore.yang
> sysrepoctl -i --yang /path/to/ietf-x509-cert-to-name.yang
> 
> sysrepocfg -d startup -i /path/to/stock_key_config.xml ietf-keystore
> mkdir -p /etc/keystored/keys
> cp /etc/ssh/ssh_host_rsa_key /etc/keystored/keys/ssh_host_rsa_key.pem
> openssl rsa -pubout -in /etc/keystored/keys/ssh_host_rsa_key.pem -out /etc/keystored/keys/ssh_host_rsa_key.pub.pem

Thanks for your time!


samuel-gauthier:
Hi Jan,

Thanks for taking the time to check this and sharing the way
you are doing cross-builds.

Maybe I miss-used 'cross-compilation' here. We are trying to
build debian packages on our side, which we are going to be
installed later on another machine.

So what we don't have when we build is actually the target
filesystem.

I think we should have separate packages for sysrepo and
netopeer2, because you can use sysrepo without netopeer2.

When the netopeer2 debian package is built, we should not call
 sysrepo binaries really, because those binaries will write in the
/etc/sysrepo directory, which "belongs to" the sysrepo package.

So I am using debian install hooks to do the installation, and netopeer2
depends on syrepo so that /etc/sysrepo is ready when netopeer2
install hooks come.

I can not share the patches, because we have our own build system
and it would not make much sense, but I am able to generate deb
packages with those two patches.

I agree that if something changes in the cmake installation
process, and we don't see it, we're screwed, but I can't think of
another way to do it... We have to be careful and follow the
changes on our side.

I am of course open to other ideas.

Samuel

jktjkt:
Ah, now I see where you're coming from. Yeah, as a former Gentoo developer, I also see the value of separate packages even though I doubt anyone would use sysrepo without netopeer2-server anytime soon.

When I was working on my patches, I was thinking about just letting CMake create a script and distributing that as a part of a device/target bootstrap. However, the current approach effectively generates a script for *CMake* which makes it unusable as-is (we don't typically have CMake on the target system, and we do not want that as a dependency). It is however possible to change the code to generate a shell script instead, and use that shell script both from CMake at install time (in the default settings) and from your post-install DEB/RPM hooks. There would be no duplication of code in that case.

michalvasko:
@samuel-gauthier
If I understand your patch correctly, it allows to skip installation of all the models and keys needed by `keystored` and manually setting `KEYSTORED_KEYS_DIR`. What you could do is not install `keystored` at all using `cmake` but simply copy the `*.so` into the correct path (that is what installation is reduced to when skipping the rest). I have no problem in adding the possibility to define `KEYSTORED_KEYS_DIR` yourself the way you did in this pull request (so I would merge that small change if in separate pull request). What do you think?

Regards,
Michal

samuel-gauthier:
Hi all,

Here is a new version, that splits the installation tasks into bash scripts that can be called or not, depending on the options given to cmake, done for keystored and server.

Regards,
Samuel

michalvasko:
Hi Samuel,
very nice, thank you!

Regards,
Michal

keystored/scripts/ssh-key-import.sh
```diff
@@ -0,0 +1,25 @@
...
+: ${CHMOD:=chmod}
+: ${OPENSSL:=openssl}
+: ${STOCK_KEY_CONFIG:=$local_path/../stock_key_config.xml}
+: ${KEYSTORED_KEYS_DIR:=/etc/keystored/keys}
+
+if [ -n "$($SYSREPOCFG -d startup --export ietf-keystore)" ]; then
+    exit 0
+fi
+
+if [ -f /etc/ssh/ssh_host_rsa_key ]; then
```
jktjkt: This is wrong because it checks that file for existence, not readability. Installing as non-root fails.

It also partially undoes changes I made there recently which let me pre-provision the configuration without actually having the key on-disk. I'll fix that one in a followup patch.

michalvasko: Well, I consider these installation scripts so you should run them as `root`. Still, there is copying performed right after it so you are right that we should check for read permissions.

samuel-gauthier: Sorry for the mistakes that were made, and thanks a lot for taking care of the fixes.

PR #264, opened 2018-05-28
SSH key provisioning: followup fixes

jktjkt:
Commit f9dafad added a method of overriding the heuristics for checking the existing openssh-style host key. This got accidentally removed in 0970cea. The use case is Buildroot and reducing the amount of stuff which has to run on the target system.

### DIFF
--- a/keystored/scripts/ssh-key-import.sh
+++ b/keystored/scripts/ssh-key-import.sh
@@ -19,10 +19,12 @@ fi
 if [ $KEYSTORED_CHECK_SSH_KEY -eq 0 ]; then
     echo "Warning: Assuming that an external script will provide the SSH key in a PEM format at \"${KEYSTORED_KEYS_DIR}/ssh_host_rsa_key.pem\"."
     $SYSREPOCFG -d startup -i ${STOCK_KEY_CONFIG} ietf-keystore
-elif [ -f /etc/ssh/ssh_host_rsa_key ]; then
+elif [ -r /etc/ssh/ssh_host_rsa_key ]; then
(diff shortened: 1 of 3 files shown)


keystored/scripts/ssh-key-import.sh
```diff
@@ -16,7 +16,10 @@ if [ -n "$($SYSREPOCFG -d startup --export ietf-keystore)" ]; then
     exit 0
 fi
 
-if [ -f /etc/ssh/ssh_host_rsa_key ]; then
+if [ $KEYSTORED_CHECK_SSH_KEY -eq 0 ]; then
+    echo "Warning: Assuming that an external script will provide the SSH key in a PEM format at \"${KEYSTORED_KEYS_DIR}/ssh_host_rsa_key.pem\"."
+    $SYSREPOCFG -d startup -i ${STOCK_KEY_CONFIG} ietf-keystore
+elif [ -f /etc/ssh/ssh_host_rsa_key ]; then
```
michalvasko: So, could you maybe change this now to `-r`?

michalvasko:
Sorry about that removed SSH key check variable, I did not notice it. Please, address the change request and it will be fine.

I would now also like to state our policy on these install scripts and any non-standard installations. We are happy to make it as easy as possible (separate install scripts should be a step forward) as long as there is not too much clutter added so that the standard installation is kept fairly straightforward.

Regards,
Michal

keystored/scripts/ssh-key-import.sh
jktjkt: Added with a warning in case the file is not readable.

michalvasko:
Alright, thanks!

Regards,
Michal

PR #265, opened 2018-05-28
Better defaults for listening configuration

jktjkt:
- enable IPv6 sockets (which include also IPv4 addresses on modern
Linux)
- use more descriptive names of the configuration
- nuke the "test" prefixes

### DIFF
--- a/server/stock_config.xml
+++ b/server/stock_config.xml
@@ -1,13 +1,13 @@
 <netconf-server xmlns="urn:ietf:params:xml:ns:yang:ietf-netconf-server">
   <listen>
     <endpoint>
-      <name>test_ssh_listen_endpt</name>
+      <name>all-interfaces</name>
       <ssh>
         <address>0.0.0.0</address>
         <port>830</port>
         <host-keys>
           <host-key>
-            <name>test_ssh_listen_key</name>
+            <name>imported SSH key</name>
             <public-key>ssh_host_rsa_key</public-key>
           </host-key>
         </host-keys>


michalvasko:
Hi,
those small cosmetic changes are fine and I have no problems with them. However, I am against changing the default interface from `0.0.0.0` to `::`. The initial implementation actually used `::` but this did not work on machines without IPv6 support so it was changed to only IPv4. It is easy to manually change it to whatever interface you want (you actually should, this default is there only so that it somehow works even out-of-the-box).

If you really want to merge this change, I must ask you to add `cmake` IPv6 support check and then change the default listening interface accordingly. Then it should work on all systems.

Regards,
Michal

jktjkt:
Hi there, sorry for disagreeing with project maintainers :), but I would like you to reconsider your position on IPv6. I think I might actually be that guy who brought up a report about a missing IPv6 support breaking stuff a long time ago. That was on a hand-crafted almost-like-Debian-but-not-quite-Debian system which was being developed at that time. We were using a custom kernel configuration which explicitly disabled IPv6 for some unknown reason (I suspect that it was designed that way back in 2009 due to some obscurity, but that's just it). But perhaps I'm wrong here -- do you remember under which contexts it broke?

However, it's 2018, we're both working for an organization which conducts research in networks, the support for IPv6 is ubiquitous, and I think that a network-oriented tool should by default listen on both IPv4 and IPv6, out-of-box.

From the technical point of view, there are two factors which affect the operation:

- Support of IPv6 in general. It's possible to compile out IPv6 altogether from the kernel. It does not matter whether a given interface has IPv6 enabled and configured or not -- it's just about the kernel (and the C library). Do you want to support systems which explicitly compile out IPv6 support? (My answer is "no", your might be different.)

- The `bindv6only` sysctl option. The Linux kernel defaults to `0` (and has **always** defaulted to zero, as far as I can tell from the git history). This means that the application does not have to do anything in order to let `::` listen on both IPv4 and IPv6. However, a system administrator or a distribution vendor can disable this, and Debian is apparently doing just that for some reason. Arguably, an application should set the [`IPV6_V6ONLY`](https://tools.ietf.org/html/rfc3493#section-5.3) socket option to make itself work reliably no matter what a kernel default is. I'll be happy to send a patch which goes in this direction if you'd like me to.

Otherwise we'll just carry a local patch for IPv6. That's also acceptable for me, it's just that I'm surprised by this default setting.

michalvasko:
Hi,
irrespective of what was the original reason for changing it to IPv4 only, we are of the opinion I presented before. We try to support also embedded devices and I think it is likely that there are perhaps some older ones that do not support IPv6.

So, I think supporting IPv6, if possible, is a good idea. It should be possible to check whether a system support it by `[ -f /proc/net/if_inet6 ]`. Will you implement this using `cmake`, please?

Regards,
Michal

jktjkt:
Fair enough. I've removed the IPv6 change for now.

I'll think about the IPv6 detection. The use case you mentioned targets deployments where the build host is different from the target embedded machine, so a simple CMake check likely won't be sufficient.

Issue #267, opened 2018-05-31
Get with selection node subtree filter returns an empty tree when the selection node is a key


alangefe:
When the selection node in a subtree filter corresponds to a node that is a key, an empty tree will be returned in a response to a get, even if there are valid matching nodes in the datastore.

I've created a unit test in alangefe/netopeer2@af946dc that demonstrates the use case and the expected behavior.

The change in behavior was introduced in 9e793f1. In prior commits this use case is working correctly. The nodes are being omitted due to the new call to `opget_build_subtree_from_sysrepo_is_key()` from `opget_build_subtree_from_sysrepo()`.

michalvasko:
Hi,
I did not thought about this use case when making the optimization, so I reverted it, thanks.

Regards,
Michal

PR #268, opened 2018-05-31
Unit test for selection node filter on key

alangefe:
This is a unit test to add test coverage for the use case in #267. It tests a selection node subtree filter whose node is a key.


michalvasko:
Hi,
thanks, there is never enough tests.

Regards,
Michal

PR #273, opened 2018-06-08
Fix session option change

olivier-matz-6wind:


michalvasko:
Hi,
of course, sorry about that. Thanks for the fix.

Regards,
Michal

PR #278, opened 2018-06-12
Fix issue where notification replay might lose notifications

alangefe:
This commit fixes a race condition in which _sysrepo_ could begin replaying notifications (in another thread) before the flag was set on the session indicating that the session could receive notifications.


michalvasko:
Hi,
you are right, thanks.

Regards,
Michal

PR #280, opened 2018-06-14
Fix missing identityref module name (fixes #279)

### LINKED ISSUE #279: Identityref predicate values do not contain the module name when the module name is the same as the identityref
### Description
With the below YANG model and snippet, the error that we am seeing is "Identityref predicate value 'melt-cdcr' missing module name". This error is being generated in resolve_partial_json_data_nodeid() in resolve.c in libyang. This problem seems to be caused by commit 310bc58 in libyang that requires an identityref value to be in the format "<module>:<value>". The root cause of this problem seems to be the following:

1) sysrepo sends just value of the identityref if the value of the identityref is in the same module as the identityref itself
2) sysrepo removes the leaf-list predicate from the xpath in rp_dt_create_xpath_for_node() in rp_dt_xpath.c

resolve_partial_json_data_nodeid() seems to get the identityref value either from the xpath or the identityref value, and neither of these has the module name, so this error is reported

This is the yang model that was used to simulate the error.
```yang
module simplified-melt {
  namespace "urn:ietf:params:xml:ns:yang:simplified-melt";
  prefix smelt;
  identity measurement-class;
  
  identity melt-cdcr {
    base measurement-class;
  }
  
  container melt {
    list pmd-profile {
      key name;
      
      leaf name {
        type string;
      }
      
      leaf-list measurement-class {
        type identityref {
          base measurement-class;
        }
        ordered-by user;
      }
    }
  }
}
```

Here is the RPC snippet.
```xml
<melt xmlns="urn:ietf:params:xml:ns:yang:simplified-melt">
  <pmd-profile>
    <name>melt-pmd-01</name>
    <measurement-class>melt-cdcr</measurement-class>
  </pmd-profile>
</melt>
```

Lastly, we noticed that some of the test code in netopeer2 that mimics sysrepo was not that same so this error could not be reproduced on netopeer2 without some code changes.

We had  to change test_edit_get_config.c to fully mimic sysrepo.
```c
int
__wrap_sr_get_item_next(sr_session_ctx_t *session, sr_val_iter_t *iter, sr_val_t **value)
{
    ...
        path = lyd_path(ietf_if_set->set.d[0]);
        /* Copied from sysrepo rp_dt_create_xpath_for_node() */
        
        /* remove leaf-list predicate */
        if (LYS_LEAFLIST & ietf_if_set->set.d[0]->schema->nodetype) {
           char *leaf_list_name = strstr(path, "[.='");
           if (NULL != leaf_list_name) {
               *leaf_list_name = 0;
           } else if (NULL != (leaf_list_name = strstr(path, "[.=\""))) {
               *leaf_list_name = 0;
           }
        }
        /* End copy */
        
        *value = calloc(1, sizeof **value);
    ...
}
```

bhart3:
Changed op_get_srval() so that it prepends the module name to the identityref value. This changes the API since op_get_srval() needs to allocate memory in this case. Therefore, a new API function was also added to free this memory, op_get_srval_free().

A new cmocka test was added that reproduces the missing identityref module error.

### DIFF
--- a/server/op_generic.c
+++ b/server/op_generic.c
@@ -33,13 +33,18 @@ build_rpc_act_from_output(struct lyd_node *rpc_act, sr_val_t *output, size_t out
 {
     struct lyd_node *node, *iter;
     uint32_t i;
-    char buf[21];
+    char *buf;
+    int alloced;
 
     for (i = 0; i < out_count; ++i) {
         ly_errno = LY_SUCCESS;
-        node = lyd_new_path(rpc_act, np2srv.ly_ctx, output[i].xpath, op_get_srval(np2srv.ly_ctx, &output[i], buf),
+        if (op_get_srval(np2srv.ly_ctx, &output[i], &buf, &alloced)) {
+            return -1;
+        }
+        node = lyd_new_path(rpc_act, np2srv.ly_ctx, output[i].xpath, buf,
                 (output->type == SR_ANYXML_T || output->type == SR_ANYDATA_T) ? LYD_ANYDATA_SXML : 0,
                 LYD_PATH_OPT_UPDATE | LYD_PATH_OPT_OUTPUT);
+        op_get_srval_free(buf, alloced);
         if (ly_errno != LY_SUCCESS) {
             return -1;
         }
--- a/server/operations.c
+++ b/server/operations.c
@@ -1837,13 +1837,22 @@ np2srv_sr_get_schema(sr_session_ctx_t *srs, const char *module_name, const char 
     return 0;
 }
 
-char *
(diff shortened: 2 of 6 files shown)


michalvasko:
Hi,
I thank you for the effort because you implemented it well but, unfortunately, I will not merge this. The problem is that when we learnt that we must always use the prefix in *libyang* and changed it accordingly, we forgot about *sysrepo*. It can either be kept the way it is (returning local identityrefs without module) and *netopeer2* needs to handle this (this pull request) or *sysrepo* must be modified to also always use itentityrefs with module prefixes. We have decided for the latter solution. Sorry.

Regards,
Michal

PR #285, opened 2018-06-18
Updating keystored logic to account for module name in addtion to identityref value

aweast:
As of https://github.com/sysrepo/sysrepo/pull/1177, we started seeing errors in sysrepo-plugin when setting the ietf-keystore algorithm parameter. 

Example snippet:
```<action xmlns="urn:ietf:params:xml:ns:yang:1">
  <keystore xmlns="urn:ietf:params:xml:ns:yang:ietf-keystore">
    <private-keys>
      <generate-private-key>
        <name>nc-server</name>
        <algorithm xmlns:ks="urn:ietf:params:xml:ns:yang:ietf-keystore">ks:rsa</algorithm>
      </generate-private-key>
    </private-keys>
  </keystore>
</action>
```
Error:
```
Generating EC private keys not supported yet.
```
After investigation, we found that after the change in https://github.com/sysrepo/sysrepo/pull/1177, sysrepo is now passing both the module name and the identityref value to sysrepo-plugin and the current code is expecting just the identityref value.

This PR should fix that behavior.

### DIFF
--- a/keystored/keystored.c
+++ b/keystored/keystored.c
@@ -298,6 +298,7 @@ ks_privkey_gen_cb(const char *UNUSED(xpath), const sr_node_t *input, const size_
     pid_t pid;
     int ret = SR_ERR_OK, status;
     char *priv_path = NULL, *pub_path = NULL, len_arg[27];
+    char *algorithm;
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are right, thanks.

Regards,
Michal

PR #294, opened 2018-07-09
when uninstall a mod,server crash!

zjbdamo:
when uninstall a mod,server crash! 
if mod-b import mod-a and only install mod-b,the mod-a will automatically installed and marked IMPORT by server. when uninstall mod-b. the mod-a is also uninstalled by server.but sysrepo will notify mod-a is unistalled. so server crash.


michalvasko:
Hi,
you are right but the message was not needed, the situation is completely valid. Thanks anyway.

Regards,
Michal

PR #311, opened 2018-08-08
Fix issue with actions from augments

alangefe:
This pull request fixes an issue with actions from augments. Before this change, an action added by an augment would always be treated as "unimplemented", even if there is a subscriber for it downstream of sysrepo.

We've written integration tests that demonstrate the problem. The fix here will allow the failing test cases (see [CI results], lines 7278-7279 and [test source code]) to pass. (We did not add a unit test for this, as main.c isn't being tested in that context right now.)

[CI results]: https://travis-ci.org/ADTRAN/netopeer2-integration-tests/builds/413288159#L7278
[test source code]: https://github.com/ADTRAN/netopeer2-integration-tests/blob/master/tests/test_action.py#L58

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -20,7 +20,7 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE debug)
 endif()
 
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi Andrew,
you are right and even though I first thought of a different solution, yours is fine, too, if not better. Thanks.

Regards,
Michal

PR #314, opened 2018-08-10
Cache ly_ctx_info to improve get performance

### LINKED ISSUE #304: <get> response times increased noticeably in netopeer2@46f2ee0
The time it takes the server to supply a response to a `get` operation has increased observably beginning in 46f2ee0.

Beginning in 46f2ee0, certain response data that used to be obtained internally by netopeer2-server (e.g., ietf-netconf-monitoring) is instead obtained from sysrepo. As a result, `get` operations are noticeably slower. The slowdown is particularly noticeable when larger numbers of models have been loaded.

Below are average `get` response times, averaged over 50 iterations, measured on the same machine, before and after 46f2ee0. The first row shows the `get` response time increase when about 70 models are present; the second row when about 300 models are present.

models | before 46f2ee0 | after 46f2ee0
--- | --- | ---
~70 | 2.2s | 6.4s
~300 | 3.3s | 9.4s

alangefe:
Fixes #304

This change results in a 3x to 4x speedup for unfiltered get requests (and for requests for the ietf-yang-library contents).

We now cache the results of the `ly_ctx_info(np2srv.ly_ctx)` call and save it in the `np2srv` global, to avoid repeating the call when it would just return the same data.

We update the cached value when necessary, namely, when a module or feature is changed.

Because the contents of the cache should change relatively infrequently, it seems worth it to pay the cost early instead of for every request.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -36,6 +36,11 @@ option(ENABLE_CONFIGURATION "Enable server configuration" ON)
 set(THREAD_COUNT 5 CACHE STRING "Number of threads accepting new sessions and handling requests")
 set(DEFAULT_HOST_KEY "/etc/ssh/ssh_host_rsa_key" CACHE STRING "Default server host key (used only if configuration is disabled)")
 
+option(ENABLE_LY_CTX_INFO_CACHE "Enable caching the ly_ctx_info() result; reduces processing at the cost of increased memory usage." ON)
+if(ENABLE_LY_CTX_INFO_CACHE)
+    set(NP2SRV_ENABLED_LY_CTX_INFO_CACHE 1)
+endif()
+
 # set prefix for the PID file
 if (NOT PIDFILE_PREFIX)
     set(PIDFILE_PREFIX "/var/run")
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi Andrew,
you are right and we have nothing against this optimization. However, I will request some changes before I will merge it.

Firstly, there is an [identificator](https://tools.ietf.org/html/rfc7895#section-2.1.1) that changes whenever the context changes so you can easily learn whether your cached info is up-to-date by comparing this value (*libyang* `ly_ctx_get_module_set_id()`), which should simplify your code.

Secondly, you certainly get much better response time for `<get>` but this data tree can be quite large so you must not forget about higher memory requirements. These are exactly the kind of optimizations we are recently working recently on, but because of the trade-off, we make them optional. So, I would like you to add a `cmake` variable `ENABLE_CACHE` which would internally translate to defining `NP2_ENABLED_CACHE` (look into *libyang*, that is what I have in mind). It can be `ON` by default.

Thank you if you implement it this way and if not, then I will do it once I have some spare time.

Regards,
Michal

alangefe:
Thanks for the feedback; I've addressed your concerns in the latest commit.

I decided to use the more descriptive flag name `ENABLE_LY_CTX_INFO_CACHE` instead of the generic `ENABLE_CACHE`, to avoid giving the impression that other data was being cached. (At the moment, we're really only caching the results from `ly_ctx_info()`.)

michalvasko:
Hi,
I asked you to name it `ENABLE_CACHE` because I expect there will be more of such optimizations in future but we can rename it then, no problem. Other than that it is OK, merging.

Regards,
Michal

PR #319, opened 2018-08-15
Fix nested containers with default children

### LINKED ISSUE #316: Containers whose leafs are set to default values are no longer suppressed in get-config response beginning in 6ab0617
Beginning in 6ab0617, the server is no longer suppressing containers whose only leafs are set to default values when it presents the configuration via a `get-config` response.

For example, given the following model:

```
container test-container {
  leaf test-leaf {
    type string;
    default "foo";
  }
}
```

and assuming the node `test-leaf` has not been set, in response to a `get-config` the server will return an empty container:

```
<rpc-reply>
  <data>
    <test-container/>
  </data>
</rpc-reply>
```

In commits prior to 6ab0617, the empty container was suppressed.

I've modified one of the test models in https://github.com/ADTRAN/Netopeer2/commit/b6d791099c1db6e33bb6f3a722e6ec8d14d9a439 to demonstrate the failure. The test `test_edit_delete1()` in `test_edit_get_config` will fail due to the unexpected empty `test-container`. The test will pass against older commits.

alangefe:
Fixes #316.

The previous fix in a9844e2 unfortunately only fixes the case where the container with the default child(ren) is at the top level of the tree. When there are containers with default child(ren) nested inside of other containers, all of the outer containers still do not have their default flag set correctly.

This is because when `lyd_insert_common()` re-evaluates the default flag, it correctly updates not only the direct parent but also all more distant ancestors:

```
    if (clrdflt) {
        /* remove the dflt flag from parents */
        for (iter = parent; iter && iter->dflt; iter = iter->parent) {
            iter->dflt = 0;
        }
    }
```

This commit uses the same idea as a9844e2, but instead of saving only the immediate parent's `dflt` flag, it counts the number of non-default ancestors. After any node is inserted, it then re-walks the parents, resetting the default flag up to the correct depth.

A better solution would probably be to modify `lyd_new_leaf()` and `lyd_new()` to take a `dflt` parameter, which could be passed from `op_sr2ly()`, and then _libyang_ would take care of keeping the default flags correct. But that would mean a _libyang_ public API change.

This commit also modifies the test model again (compare #317) to have multiple nested children, and verifies the new fix.

### DIFF
--- a/server/operations.c
+++ b/server/operations.c
@@ -2822,8 +2822,8 @@ op_sr2ly(struct lyd_node *root, const sr_val_t *sr_val, struct lyd_node **new_no
 {
     char numstr[22];
     char *val_str;
-    int new_node_i = -1, parent_dflt;
-    struct lyd_node *parent = NULL, *node;
+    int new_node_i = -1, parent_dflt_depth = 0;
+    struct lyd_node *parent = NULL, *node, *iter;
     const struct lys_module *mod = NULL;
 
     /* last used index in cache is ours, we can learn everything based on that */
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi Andrew,
yeah, this looks good. This is unfortunately the case, when a function is trying to do all the work but it is actually counterproductive, but I think it is just fine how you changed it. Thanks.

Regards,
Michal

PR #328, opened 2018-08-27
Provide the other certs from the server's cert chain

alangefe:
This commit from @jwwilcox, together with a corresponding commit to _libnetconf2_, fixes the TLS connection scenario in which the server's certificate has been signed by an intermediate CA, but the client only has the root CA available locally. In this case, the client will reject the connection attempt, because it does not know about the intermediate CA.

This change adds an optional callback so that the server can supply the missing certificate. The changes in _libnetconf2_ will use the certificates retrieved from this callback to allow the server's TLS context to automatically provide the intermediate certificate(s) to the client.

This scenario is demonstrated in the integration test `test_tls_client_missing_server_intermediate()` in [ADTRAN:netopeer2-integration-tests](https://github.com/ADTRAN/netopeer2-integration-tests/blob/master/tests/test_tls.py#L73). The changes here, together with the corresponding commit in _libnetconf2_, will allow [the currently failing test case](https://travis-ci.org/ADTRAN/netopeer2-integration-tests/jobs/420293391#L7434) to pass.

### DIFF
--- a/server/ietf_keystore.c
+++ b/server/ietf_keystore.c
@@ -102,6 +102,57 @@ np_server_cert_clb(const char *name, void *UNUSED(user_data), char **UNUSED(cert
 }
 
 int
+np_server_cert_chain_clb(const char *name, void *UNUSED(user_data), char ***UNUSED(cert_paths), int *UNUSED(cert_path_count),
+                         char ***cert_data, int *cert_data_count)
+{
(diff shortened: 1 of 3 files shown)


alangefe:
This commit goes with https://github.com/CESNET/libnetconf2/pull/76.

PR #332, opened 2018-08-31
Fix list without key

alangefe:
This commit fixes the scenario in which the server needs to report a list without keys. (This is valid for lists that do not represent configuration; see [RFC 7950 section 7.8.2.](https://tools.ietf.org/html/rfc7950#section-7.8.2)) This scenario was broken by the changes in e75e339, which caused `op_sr2ly()` to fail if the list's xpath predicate did not contain a `child="value"` expression.

I made two modifications to `op_sr2ly_parse_pred()`:
1. Fix the loop that searches for `'='` so that it stops at the NULL termination of the string if no `'='` has yet been found. (The existing code could search far beyond the bounds of the string, sometimes resulting in a segmentation fault.)
2. Distinguish between a failure to find a `child="value"` predicate, which is a valid case, and other general parsing errors, which are not.

I then changed `op_sr2ly_create_keys()` so that it no longer fails if no list keys are found.

I also added a test case that demonstrates the problem. I simulated the `netconf-config-change` notification from [RFC 6470](https://tools.ietf.org/html/rfc6470), which contains a list without a key. The test case will pass in commits prior to e75e339, fail with e75e339, and pass again with this patch applied.

### DIFF
--- a/server/operations.c
+++ b/server/operations.c
@@ -2631,11 +2631,18 @@ op_filter_create(struct lyd_node *filter_node, char ***filters, int *filter_coun
     return 0;
 }
 
+enum {
+    OP_SR2LY_PARSE_PRED_SUCCESS = 0,
+    OP_SR2LY_PARSE_PRED_NO_KEY,
+    OP_SR2LY_PARSE_PRED_PARSE_ERROR
+};
+
 static int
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
yeah, I have not covered this use-case at all, thanks!

Regards,
Michal

PR #350, opened 2018-11-12
Update syslog message priority format to match sysrepo

bhart3:
Changes for syslog priority strings which was done for consistency to match the the sysrepo priority format in syslog messages.  The goal is to allow external filtering on combined syslog files.


michalvasko:
Alright, merging.

Regards,
Michal

PR #362, opened 2018-12-07
Update README.md

tehelsper:
We'd like to build awareness of the netopeer integration test suite we have built by adding a build status icon to the README for Netopeer2. This automatically builds all of the upstream NETCONF packages together and runs a series of tests against a fully running system that we have found very useful in the past. It currently runs once per day. If you are open to it, we could modify it to run automatically for each commit.


rkrejci:
Thanks Taylor, adding the link into README is no problem, I welcome it. But actually I would probably prefer to integrate the tests themselves (so the whole netopeer2-integration-tests repository) into the Netopeer2 repository and run it in Travis with each push. Or do you have some reason to keep the tests in the separated repository?

tehelsper:
Radek, that sounds great! We'll work on getting that integrated and submit another PR.

PR #372, opened 2018-12-21
Update netopeer2 to run integration tests on check in

bhart3:
We have worked on adding the integration tests into the netopeer2 build directly as we discussed in PR #362. This should add a new makefile target to run the integration tests, and we have added that to the travis script.

Everything under integration-tests was pulled directly from the latest of the [Netopeer2-integration-tests](https://github.com/ADTRAN/netopeer2-integration-tests) git repo.


.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: c
 
+sudo: required
```
bhart3: This was added since the integration tests run in a docker container and this requires sudo to run a docker container.

Issue #363, opened 2018-12-10
New feature


wmakuch:
Hello Michal!

Netopeer2 and sysrepo are great tools used in my project for communication via netconf.
There are some features I need to use, but there is no support provided.
I would like to open discussion and ask you about advices, code review and opinion about feature I prepared.
The feature is:
"Application should receive information from which netconf client (session-id) specific rpc has been sent."

My questions:
1. Would you like to merge such feature into repository (Netopeer2 and sysrepo)?
2. What is the best way to write feature like this? - I just added field in sysrepo.proto and extended rpc_cb with additional parameter.
3. I cannot dereference struct nc_session* and get id without copy entire class from netconf into Netopeer2. What is best way to avoid it?
In the future I am going to extend this feature: print session-id in notification, get info when rpc has been sent.

Please let me know, if you wish to see my changes.
If yes - in what form should I send it - you prefer patches? or open pull request in Netopeer2 and sysrepo repository?

Best Regards,
Wojciech Makuch

michalvasko:
Hi,
we are glad you are using our NETCONF suite and we generally always welcome patches, best in the form of a pull request.

1. Create a pull request and we will see, probably yes.
2. Seems OK but I would have to see the changes themselves to be sure.
3. Using `nc_session_get_id()`?

Regards,
Michal

wmakuch:
Hello,

Thank you for quick reply!
Soon I will open new pull request.

Best Regards,
Wojciech Makuch

PR #376, opened 2019-01-02
Application receive information from which netconf client (session-id…) specific rpc has been sent.

wmakuch:
**Description**
Extended _rpc_cb_ with parameter _netconf_id._ This value is set in Netopeer2-server. In the case of unknown netconf id, this parameter should be 0.

**Test case**
Manual tests.

**Closure**
Netopeer2 #363

### DIFF
--- a/server/op_generic.c
+++ b/server/op_generic.c
@@ -170,9 +170,9 @@ op_generic(struct lyd_node *rpc, struct nc_session *ncs)
     }
 
     if (!act) {
-        rc = np2srv_sr_rpc_send(sessions->srs, rpc_xpath, input, in_idx, &output, &out_count, &ereply);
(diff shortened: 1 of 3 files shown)


wmakuch:
Hi,

I updated pull request.
After further testing I realised that API must be extended in order to make this feature useful.
Your suggestion was pass `sr_session_ctx_t*` by `private_ctx`, but this parameter is set during `sr_rpc_subscribe` (not during `sr_rpc_send`), so _Netopeer2-server_ has nothing to do with it. I decided to pass `sr_session_ctx_t*` directly.

Kindly please take a look at my changes and share your opinion, comments.

PR to _Netopper2_ repository: https://github.com/CESNET/Netopeer2/pull/376

Best Regards,
Wojciech Makuch

PR #386, opened 2019-01-29
Fix two errors in <copy-config> to URL

kpbarrett:
- Need WITHSIBLINGS, output only had one module included
 - Was copying config=false nodes


michalvasko:
Hi,
seems good to me, thanks.

Regards,
Michal

PR #406, opened 2019-03-11
close memory leak (test case + solution)

frimpler:
A memory leak reg. nc-poll-sessions has been detected during shutting down netopeer2-server. This PR provides test case and solution.

Regards,
Frank

### DIFF
--- a/server/main.c
+++ b/server/main.c
@@ -1772,6 +1772,11 @@ cleanup:
         sr_unsubscribe(np2srv.sr_sess.srs, np2srv.sr_subscr);
     }
 
+    /* close all open sessions */
+    while (nc_ps_session_count(np2srv.nc_ps)!=0) {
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
seems fine but with regards to the last changes, please also use the provided `ATOMIC_UINT32_T` for your counter as it is set/accessed from several threads. Thanks.

Regards,
Michal

frimpler:
Please have a look, again - hope, that's ok that way.

michalvasko:
Hi Frank,
I have changed those atomic macros a bit, again. To compile, you only need to rename `ATOMIC_UINT32_T` to just `ATOMIC_T`. However, please also use `ATOMIC_LOAD` and `ATOMIC_STORE` wherever needed. Thanks.

Regards,
Michal

frimpler:
adapted it...

michalvasko:
Hi Frank,
okay, thanks.

Regards,
Michal

PR #408, opened 2019-03-12
tests: Enable parallel test execution

jktjkt:
All tests were trying to use the same PID file, and therefore they would fail to start properly and eventually hit a timeout, outputting an error:
```
 [ RUN      ] test_startstop
 [ERR]: Another instance of the Netopeer2 server is running.
```
I cannot use `__FILE__` directly because that one might contain full path. There's `__builtin_strchr`, but doing this via CMake is probably a bit easier.

Also adjust the temporary file name for XML operations to be consistent.


michalvasko:
Hi,
the changes seems fine, merged.

Regards,
Michal

PR #412, opened 2019-03-22
[🐛 FIX] segmentation fault on CLI when using pipe

andrei-pavel:
Netopeer2 CLI 2.0.51
Compile time: Mar 22 2019, 15:55:56

Using pipe as input to `netopeer2-cli` is an easy way to automate some netconf commands.
But it crashes on duplicate commands. Sometimes. It might depend on the contents of history that is dumped to the file at the time that you start `netopeer2-cli`. There have been times where this didn't replicate for me.

netopeer2-cli segmentation fault

If it were a double free, checking for null would have fixed it.
It rather seems to be a stack corruption.
Looked at the address of `ls.history[ls.history_len-1].data` one time and it was `0x21`.

### DIFF
--- a/cli/linenoise/linenoise.c
+++ b/cli/linenoise/linenoise.c
@@ -325,7 +325,7 @@ static void freeCompletions(linenoiseCompletions *lc) {
         free(lc->cvec);
 }
 
-/* This is an helper function for linenoiseEdit() and is called when the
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
please try it now, your changes would cause memory leaks and not fix the underlying problem anyway.

Regards,
Michal

PR #440, opened 2019-06-26
Switch X509_get_notAfter to get0 variant

neheb:
The former is deprecated.


michalvasko:
Hi,
okay, thanks.

Regards,
Michal

Issue #450, opened 2019-08-02
test_get fails: 


jktjkt:
The `test_get` fails on my system and within our internal CI. I'm upgrading to the following versions:
- CESNET/libyang@3a8aeb6fac7a55ff50f9acd81d3b939c2127f9cc
- sysrepo/sysrepo@480a0cd5d17a8dfc0390cd03f7fd9bc821b0e218
- CESNET/libnetconf2@2c63a6df826cf882f03bd0a235fc105f60752d93
- Netopeer2 commit CESNET/Netopeer2@4aeaaeca3a96cd6136a1a924b86e4d150e75e4e2

This is the test failure:

```
$ ctest -R test_get --output-on-failure
Test project /home/jkt/work/prog/_build/server/GCC_8_1_NETCONF-Debug
    Start 2: test_get
1/1 Test #2: test_get .........................***Exception: Child aborted  0.36 sec
[==========] Running 6 test(s).
[ RUN      ] test_startstop
[INF]: Netopeer2 connected to sysrepod.
[INF]: Plugin "/opt/nc/lib64/libyang/extensions/nacm.so" successfully loaded.
[INF]: Plugin "/opt/nc/lib64/libyang/extensions/metadata.so" successfully loaded.
[INF]: Plugin "/opt/nc/lib64/libyang/extensions/yangdata.so" successfully loaded.
[INF]: Plugin "/opt/nc/lib64/libyang/user_types/user_yang_types.so" successfully loaded.
[INF]: Plugin "/opt/nc/lib64/libyang/user_types/user_inet_types.so" successfully loaded.
[INF]: Reading module "ietf-yang-metadata".
[INF]: Module "ietf-yang-metadata@2016-08-05" successfully parsed as implemented.
[INF]: Reading module "yang".
[INF]: Resolving "yang" unresolved schema nodes and their constraints...
[INF]: All "yang" schema nodes and constraints resolved.
[INF]: Module "yang@2017-02-20" successfully parsed as implemented.
[INF]: Reading module "ietf-inet-types".
[INF]: Resolving derived type "union" failed, it will be attempted later.
[INF]: Resolving derived type "union" failed, it will be attempted later.
[INF]: Resolving derived type "union" failed, it will be attempted later.
[INF]: Resolving derived type "union" failed, it will be attempted later.
[INF]: Resolving "ietf-inet-types" unresolved schema nodes and their constraints...
[INF]: All "ietf-inet-types" schema nodes and constraints resolved.
[...]

PR #451, opened 2019-08-02
Adapt to ietf-yang-library update within libyang

jktjkt:
It seems that libyang now uses RFC 8525, so let's adapt the test suite:

- features are reported in a different way
- the "checksum" got renamed to "content-id"
- there's no checksum within a module-set anymore

Fixes: CESNET/libyang@134fa341 ("libyang CHANGE updated ietf-yang-library revision")
Fixes: https://github.com/CESNET/Netopeer2/issues/450


michalvasko:
Hi Jan,
sorry about that and thanks for the fix.

Regards,
Michal

PR #461, opened 2019-09-05
netconf-monitoring segv fix

kpbarrett:
This change fixes a segfault with non-SSH/TLS sessions, where netconf monitoring stats calls are being called for these sessions that are not in the list of sessions. I believe the stats calls are still being called from np2srv_ntf_send().


michalvasko:
Hi,
you are right but I would prefer a fix similar to [this](https://github.com/CESNET/Netopeer2/blob/master/server/main.c#L766) rather than just ignoring if the session is not found.

Regards,
Michal

PR #462, opened 2019-09-08
Fix for only SSH or TLS enabled

brchiu:
When only NC_ENABLED_SSH defined, NC_TI_OPENSSL does not exist in
enumeration type NC_TRANSPORT_IMPL. Similarly, when only NC_ENABLED_TLS
defined, NC_TI_LIBSSH does not exist in NC_TRANSPORT_IMPL.

So we have to re-factor this code excerpt for compilation.


michalvasko:
Hi,
seems correct, thanks.

Regards,
Michal

PR #467, opened 2019-10-03
keystored: force public key in pem format

rjarry:
Fix the following error:

```
[ERR] File "/etc/keystored/keys/ssh_host_rsa_key.pem.pub" is not a public key in PEM format.
```

ssh-keygen ignores the `-m pem` argument for the public key file. It always generates it in the openssh format.

Convert it to PEM afterwards.

`ssh-keygen` requires that the converted file has `0600` permissions (it assumes it is a private key). Restrict permissions before converting and restore them afterwards.


michalvasko:
Hi,
okay, it seems you are right, thanks.

Regards,
Michal

rjarry:
Thanks

PR #470, opened 2019-10-04
server: listen on ipv4 & ipv6 by default

rjarry:
Do not only listen on ipv4. Some people have ipv6 networks.

### DIFF
--- a/server/stock_config.xml
+++ b/server/stock_config.xml
@@ -3,7 +3,7 @@
     <endpoint>
       <name>all-interfaces</name>
       <ssh>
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
it is intended that only IPv4 interfaces are listened on for systems that do **not** support IPv6 (it was discussed some time ago, originally it was how you propose it). If you do, change it, this is just the default.

Regards,
Michal

rjarry:
Ok I understand. I'll change it on my end.

Issue #548, opened 2020-02-05
get-data and edit-data CLI commands return the wrong help


bbergquist0930:
These commands return the get-config and edit-config command help respectively.  When fixed, the get-config help will cause a segmentation violation if there is no session.

bbergquist0930:
Works with the latest code now.

PR #549, opened 2020-02-05
Fixed an issue where the wrong help is being displayed

bbergquist0930:
This is for issue

https://github.com/CESNET/Netopeer2/issues/548

for the 'get-data' and 'edit-data' commands.

Fixed an issue where if there is no session established
and you invoke a 'get-data' command a segmenation violation
occurs because of accessing an unitialized stack pointer
variable.


cli/commands.c
```diff
@@ -1082,7 +1082,7 @@ cmd_getschema_help(void)
 void
 cmd_getdata_help(void)
 {
-    const struct lys_module *mod;
+    const struct lys_module *mod = 0;
```
michalvasko: Please change it to `NULL`, it is a pointer.

PR #566, opened 2020-02-26
Added compile time variable for setting RPC timeout

sriramkandukuri:


server/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ set(THREAD_COUNT 5 CACHE STRING "Number of threads accepting new sessions and ha
 set(MODULES_OWNER "$ENV{USER}" CACHE STRING "User that will become the owner of server modules")
 set(NACM_RECOVERY_UID 0 CACHE STRING "NACM recovery session UID that has unrestricted access")
 set(DATA_CHANGE_TIMEOUT 0 CACHE STRING "Timeout in seconds of any (sysrepo) data changes, 0 is the default sysrepo timeout")
+set(RPC_TIMEOUT 0 CACHE STRING "Timeout in seconds of any RPC, 0 is the default sysrepo timeout")
```
michalvasko: Please set the default to 6 seconds. The description should explain that this is the timeout of the global RPC callback that calls all the other specific RPC callbacks meaning this timeout should always be higher than the maximum timeout of the other RPCs. Timeout of the other RPCs boils down to the timeout of sysrepo API calls in those callbacks. The highest default timeout (which can be changed by `DATA_CHANGE_TIMEOUT`) is 5 seconds, hence this being 6 seconds.

sriramkandukuri: Done please check.

PR #593, opened 2020-04-16
Sync ietf-datastores with the version provided by libyang

jktjkt:


jktjkt:
Here's how it's failing with the most recent libyang: https://zuul.gerrit.cesnet.cz/t/public/build/f59ad09409ab4320b7cc3b91d5f49924

michalvasko:
Hi,
right, sorry. Although I do not think require the latest *libyang* in this old server is desired so the solution here is probably to remove the test that fails.

Regards,
Michal

jktjkt:
I think that those who cannot update to latest libyang can just stay on an oldish Netopeer2. I've deliberately bumped the libyang requirements to the latest because that's what I'm testing against, and I don't think that a full bisection run for finding out which version exactly introduced this would be worth the hassle.

michalvasko:
I agree, that is why I suggested removing the single test altogether, which should make it, again, not dependent on any *libyang* version.

jktjkt:
I think that the test is reasonable because it can guard against possible future regressions in libyang. It should be low-maintenance because I do not expect changes to that IETF model now that it's been released. I understand that your interest in maintenance of legacy code is not big :), but this is a patch which solves a real problem. (And yup, we will eventually move to the non-legacy branches, it's just `ENOTIME` for now.)

michalvasko:
Well, ultimately I can fix it any way I want but I do not really understand why you think your change is the most correct one. The test itself is a rather useless one considering there are filtered tests following it and, most importantly, requires an exact answer from the server, which, as happened now, is a maintenance problem. It depends on an exact set of internal *libyang* modules and their revisions, which may easily change in future. Also, like I said before, this change now requires the latest *libyang*, which is an overkill for such legacy code.

So, I will simply remove that one test to fix this unless you can give me a really good reason why I should not.

Regards,
Michal

jktjkt:
I missed the fact that there are more fine-grained tests for other libs in the library, so I agree with you, this looks like something that can be safely removed.

michalvasko:
Okay, should work now without the test.

Regards,
Michal

Issue #598, opened 2020-04-20
Netopeer2 support on QNX platform


apropp-molex:
Hi Michal,

So I see that there are a couple of issues open pertaining to QNX, but none really went anywhere. As usual, I've summarized my issues below. Let me know if you need more details or anything else. Thanks for your help thus far.

In terms of both `cli` and `server` I ran into some issues:
1. Manually specifying `-DLIBSSH_LIBRARY` and `-DLIBSSH_INCLUDE_DIR` had no effect, unlike with `libnetconf2`. This caused it to be perceived that `libnetconf2` wasn't compiled with ssh support when it was really a missing header issue.
1. Both needed in some form the usual GNU/non-QNX functions: `asprintf`, `getline`, `vasprintf` and `strndup`. This issue has been dealt with for `libyang`, `sysrepo` and `libnetconf2`. Would it be possible to do here as well?
1. Due to issue 1, obviously linking with the `-lssh` flag failed.

The `cli` had a few other functions missing that were GNU extensions:
1. `sendfile` - could be done with `read/write`
1. `strchrnul` - `strchr` is on QNX
1. `strdupa` - `strdup` and `alloca` are on QNX as well

With the server, as I couldn't specify where to look for `libssh` files, it opted to link against `-lssh_threads`. Given that I'm pulling from the top of master, that obviously won't link. `-lcrypt` also tried to link but as discussed in `libnetconf2` this is implemented with `-llogin` on QNX.

michalvasko:
Hi,
so please try the current `devel-server`, most of the issues should be fixed. Actually, I am hoping that all of them are except for the specific `netopeer2-cli` missing functions. Please provide a pull request for that so that `sendfile` is not used at all and `strchrnul` and `strdupa` are added to the *compat* used in *netopeer2* with all needed checks.

Regards,
Michal

apropp-molex:
Hi Michal,

Tried the current `devel-server`. `netopeer2-server` builds fine. `netopeer2-cli` still doesn't make use of `-DLIBSSH_LIBRARY` and `-DLIBSSH_INCLUDE_DIR`. This leads to issues with finding header files and the `.so` files. Any thoughts on this?
Pull request has been provided for the missing functions.

michalvasko:
Hi,
you have not mentioned CLI before so I just assumed the old `FindLibSSH` from `master` was causing problems. Also, why should the CLI be linked with *libssh*? It should not be using any functions from it, I have checked just now.

[...]

PR #600, opened 2020-04-21
netopeer2 CHANGE added more compat functions

apropp-molex:
Refs #598
`strdupa`, `strchrnul` and reimplemented `cp` function in `cli/commands.c` without `sendfile`.


michalvasko:
Hi,
right, I did not even realize that `strdupa` will have to be a macro. But looks good, merging.

Regards,
Michal

PR #609, opened 2020-04-29
build CHANGE support libyang custom location

dumitrupaul:
Hello,

I had an issue while trying to use cmake with libyang installed in a custom location because `check_symbol_exists("NC_ENABLED_SSH" "nc_client.h" LIBNETCONF2_ENABLED_SSH)`
`check_symbol_exists("NC_ENABLED_TLS" "nc_client.h" LIBNETCONF2_ENABLED_TLS)`
would fail with a 'file not found error'.
This fixes it.

Kind regards,
Paul.


michalvasko:
Hi,
I cannot really merge this because it expects that *libyang* is searched before *libnetconf2* and in general this does not seems like a solution, rather a workaround at best. Also, your patch duplicates [existing code](https://github.com/CESNET/Netopeer2/blob/devel-server/server/CMakeLists.txt#L163) so I do not really understand why it fixes anything. Could you please investigate more?

Regards,
Michal

andrei-pavel:
Hi,

This is a fix for an error that has impacted me as well consistently for two years now. CMake reports that SSH support is not enabled in libyang when in fact it is. The technical problem is that the `check_symbol_exists()` directive looks for the define symbols, but fails to validate the `nc_client.h` header itself because the header include chain goes down to the `libyang.h` header which it cannot find because it doesn't look in the custom location of libyang. Here is an excerpt from the CMakeError.log that @dumitrupaul actually found:
```
In file included from /opt/libnetconf2/include/nc_client.h:32:
/opt/libnetconf2/include/libnetconf2/session_client.h:22:10: fatal error: 'libyang/libyang.h' file not found
#include <libyang/libyang.h>
```
The way I would solve this is just remove those checks. And it would compile just fine with SSH support. But it is obviously not the way to go.

It's easy to replicate, remove any system-wide installation, install in `/opt/libyang`, point sysrepo to `/opt/libyang`. Could you check?

I don't know why the existing line doesn't solve it, maybe the path resets on `find_package(LibNETCONF2 REQUIRED)`

I guess most people always install in the default prefix or set C_INCLUDE_PATH or run in docker, I don't understand how this flew by unnoticed.

Best regards!

michalvasko:
Hi,
okay, I have checked this and it worked fine for me on Ubuntu 18.04. I installed *libyang*/*libnetconf2*/*sysrepo* into `/usr/local1` and so I had to always manually specify all the paths. But when I did, *netopeer2* configured and compiled just fine. So, you will have to provide more details if you want this fixed.

Regards,
Michal

andrei-pavel:
Hi,
Your dockerfile[[1]](https://github.com/sysrepo/sysrepo/blob/master/deploy/docker/sysrepo-netopeer2/Dockerfile.devel) works.
Dockerfile modified to install in custom paths[[2]](https://github.com/andrei-pavel/showcase/blob/master/sysrepo/Dockerfile-opt) fails with:
```
CMake Error at CMakeLists.txt:56 (message):
  Missing SSH and TLS support in libnetconf2, CLI requires at least one them
  to be supported.
```
```
docker build -t sysrepo -f ./Dockerfile-opt .
```

[1] https://github.com/sysrepo/sysrepo/blob/master/deploy/docker/sysrepo-netopeer2/Dockerfile.devel
[2] https://github.com/andrei-pavel/showcase/blob/master/sysrepo/Dockerfile-opt

andrei-pavel:
If you want to further replicate locally, I suggest removing other libyang, sysrepo installs and unsetting `C_INCLUDE_PATH` & `CPP_INCLUDE_PATH`.

michalvasko:
Hi,
so I have added another explicit check that caused printing an inaccurate message in my case. But it worked the way it should have even on another system I tried it on (OpenSUSE Tumbleweed). So, please provide __exact__ `cmake` commands for *libyang*, *sysrepo*, *libnetconf2*, and finally *netopeer2-server* so that I have a chance of reproducing this. `docker` cannot really be used for this because in my experience it often executes different commands than I expect so I am not able to debug anything with it.

Also, I will rewrite our `cmake` find modules a bit soon to be compliant with all the standard ones but I do not think it will fix this problem.

Regards,
Michal

andrei-pavel:
Hi,
The commands from the Docker are exactly what we use locally. Even the `export LD_LIBRARY_PATH` is used locally to get over a `SYSREPOCTL-NOT_FOUND` in `/opt/dev/Netopeer2/setup/install.sh` just like in docker. We'll maybe come back to that later. Please try those. We are using Ubuntu 18.04 and 20.04.

I'm aware that I give the cli some extra arguments that are not used, but that's only so that they are the same as for the server and it shouldn't impact the issue.
Regards!

michalvasko:
Okay, I have found out what the problem was. Not sure what issue exactly is this pull request solving but it should have been explicitly said that it is only the __CLI__ that fails to be configured, the __server__ works fine. I have been testing it only with the server. So, few hours wasted and I will not even fix it because CLI `cmake` list will be a sublist of the main, server one. Meaning all the defined variables will simply be reused.

Regards,
Michal

Issue #628, opened 2020-05-26
sr_get_data error details are not returned in netconf replies


rjarry:
Hi Michal,

When calling `sr_get_data()` directly and one of the operational data callbacks (from an application) returns an error with additional information (reported via `sr_set_error()`), `sr_get_error()` returns the error details that were reported in the application callbacks.

However, when sending a `/ietf-netconf:get` rpc to netopeer2-server and one of the application callbacks return an error, the error reply only contains a generic error message: `User callback failed`. The detailed error message returned by the application is lost in translation.

I tried to fix the problem myself, but I don't know where to start. I hope you can help :)

Thanks in advance.

rjarry:
Note that other RPC errors details seem to be forwarded properly. The issue only affects `/ietf-netconf:get`.

rjarry:
Actually, I have a patch that fixes the problem:

```diff
diff --git a/src/netconf.c b/src/netconf.c
index 9cf9db72a016..70d165567b0e 100755
--- a/src/netconf.c
+++ b/src/netconf.c
@@ -98,7 +98,15 @@ get_sr_data:
     for (i = 0; i < filter_count; i++) {
         rc = sr_get_data(session, filters[i], 0, 0, get_opts, &node);
         if (rc != SR_ERR_OK) {
+            const sr_error_info_t *err_info = NULL;
             ERR("Getting data \"%s\" from sysrepo failed (%s).", filters[i], sr_strerror(rc));
+            if (sr_get_error(session, &err_info) == SR_ERR_OK && err_info != NULL) {
+                int e;
+                for (e = 0; e < err_info->err_count; e++) {
+                    sr_set_error(session, err_info->err[e].xpath,
+                                 "%s", err_info->err[e].message);
+                }
+            }
             goto cleanup;
         }
 ```

Does that look ok for you?

PR #631, opened 2020-05-27
netconf BUGFIX forward get_data errors to client

rjarry:
When error details are set by the client application that provide the requested data, forward them to the netconf client.

Fixes: #628


michalvasko:
Hi,
while you can set several custom errors, only the first one is actually sent back to the originator. So I would expect the same code as in all the [other](https://github.com/CESNET/netopeer2/blob/devel/src/netconf.c#L234) RPCs. Not sure why I did not put it into this single one.

Regards,
Michal

rjarry:
Hi Michal, it should be ok now.

michalvasko:
Yep, thanks.

Regards,
Michal

PR #632, opened 2020-05-27
netconf BUGFIX use data change timeout for sr_get_data

rjarry:
All sr_get_data calls use the default sysrepo timeout, NP2SRV_DATA_CHANGE_TIMEOUT is not taken into account.

Use NP2SRV_DATA_CHANGE_TIMEOUT for all sr_get_data() calls.


michalvasko:
Hi,
okay, it makes sense except I am also going to rename it because this use-case does not really fit its name nor description.

Regards,
Michal

PR #633, opened 2020-05-27
server BUGFIX do not abort endpoint changes processing on error

rjarry:
When changing the configuration to add new listen endpoints, if one of these endpoints specifies an invalid parameter (for example, a "local-address" which is not present on any interface on the system), the nc_server_endpt_set_address() function will return an error (since the socket cannot be created).

If such errors occur, do not abort the processing of changes since other endpoint address/port may be valid and should be configured. Otherwise, these changes are lost and will never be configured.

If one error occurred, return an error to notify the sysrepo client which applied the configuration that something went wrong.

Signed-off-by: Robin Jarry <robin.jarry@6wind.com>

### DIFF
--- a/src/netconf_server.c
+++ b/src/netconf_server.c
@@ -171,7 +171,7 @@ np2srv_endpt_tcp_params_cb(sr_session_ctx_t *session, const char *UNUSED(module_
     const char *prev_val, *prev_list, *endpt_name;
     char *xpath2;
     bool prev_dflt;
-    int rc;
+    int rc, failed = 0;
 
     if (asprintf(&xpath2, "%s/*", xpath) == -1) {
         EMEM;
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi Robin,
actually, the proper way for this to work would be to utilize the *sysrepo* functionality for this meaning make all the changes happen on `SR_EV_CHANGE` instead of `SR_EV_DONE`. But then even `SR_EV_ABORT` would have to be handled everywhere (which is probably why I implemented it in this simplified way) for the rollback to occur.

In NETCONF terms, it now behaves as `stop-on-error` whereas your change makes the behaviour as `continue-on-error`. Neither solutions are good because you will always end up with configuration that is not applied but I guess with this change more of it is applied than currently. So, I do not really have anything against it. Any thoughts?

Regards,
Michal

rjarry:
Yes I agree that the 100% correct approach would be to validate that the changes will be applied successfully on `SR_EV_CHANGE`. However that would require more extensive code modifications which I did not want to make to avoid breaking something else.

Note that there are some module change callbacks which I did not modify meaning they are still behaving as `stop-on-error`.

Maybe this is acceptable for the time being.

michalvasko:
Right, I suppose so. When the used modules are finally finished and released as RFCs, I may implement it properly, we will see. However, note that any return value from `SR_EV_DONE` event is completely ignored so it would probably be best if the code reflected that meaning you can simply ignore return values of all the *libnetconf2* functions.

Regards,
Michal

rjarry:
I can ignore all return values but it will cause compiler warnings. Bypassing these warnings is quite tedious as a naive `(void)` cast on the function call does not silence them. The only way is to assign the result to a variable and explicitly ignore it.

michalvasko:
Okay then, merged.

Regards,
Michal

Issue #636, opened 2020-06-03
Moving leaf list instances to "last" doesn't work


syyyr:
Hello, I'm trying to move user-ordered leaf list instances around using netconf. Moving to "first" works for me, but moving to "last" doesn't work.
I have this schema:
```yang
module example-schema {
    prefix aha;
    namespace "http://example.com";

    leaf-list protocols {
        type string;
        ordered-by user;
    }
}
```
This is the code I tried:
```cpp
#include <libyang/Libyang.hpp>
#include <iostream>
#include <libyang/Tree_Data.hpp>
#include <sysrepo-cpp/Session.hpp>
extern "C" {
#include <nc_client.h>
}


int main(int argc, char* argv[])
{
    auto sysrepo_connection = std::make_shared<sysrepo::Connection>("test");
    auto sysrepo_session = std::make_shared<sysrepo::Session>(sysrepo_connection);
    auto sysrepo_callback = std::make_shared<sysrepo::Callback>();
    auto sysrepo_subscription = std::make_shared<sysrepo::Subscribe>(sysrepo_session);
    sysrepo_subscription->module_change_subscribe("example-schema", sysrepo_callback);

    auto ctx = std::make_shared<libyang::Context>();
    ctx->parse_module_path("./example-schema.yang", LYS_IN_YANG);
    auto session = nc_connect_unix("/home/vk/git/netconf-cli/build/netopeer2-server.sock", nullptr);

    auto printData = [&session] {
        auto rpc = nc_rpc_get("/example-schema:protocols", NC_WD_ALL,  NC_PARAMTYPE_CONST);
        uint64_t msgId;
        nc_send_rpc(session, rpc, -1, &msgId);
        struct nc_reply* reply;
        nc_recv_reply(session, rpc, msgId, -1, 0, &reply);
        if (reply->type == NC_RPL_DATA) {
            nc_reply_data* reply_data = reinterpret_cast<struct nc_reply_data*>(reply);
            auto data = libyang::create_new_Data_Node(reply_data->data);
            if (!data) {
                std::cout << "got empty data" << std::endl;
                return;
            }
            std::cout << "printing data:\n" << data->print_mem(LYD_XML, LYP_WITHSIBLINGS) << "\n";
        } else {
            std::cout << "nc_rpc_get: error\n";
        }
    };

    auto createItem = [&ctx, &session] (auto path) {
[...]

PR #639, opened 2020-06-03
Fix list and leaf-list moves

jktjkt:
Two fixes for issues in moving of `list` and `leaf-list` YANG types (#636 by @syyyr). I don't know whether the non-legacy branch also has these bugs (I have not found any tests for moves neither in `legacy`, nor in `master`).

## Allow moving list/leaf-list items to the last position

The code wanted to deliberately ignore any `SR_MOVE_LAST` values. I think that this happened because of that C convention for adding a `FOO_LAST` items to enums to signify their largest allowed value.
    
I did not want to modify sysrepo for this as well, so I just added a random garbage sentinel for this.
    
## Do not reset relative list XPath too early

When processing YANG list moves, the code would correctly parse the relative XPath in the `edit_get_move()`, but before it could be consumed via `np2srv_sr_move_item`, it got reset when the main iteration loop processes the next change item. As a result, a relative move operation was always attempted with a `nullptr` for the relative item XPath.
    
Fix this by only resetting the `rel` when encountering a new `leaf-list` or a new `list`. I have no clue if this is actually correct (what about nested lists, for example?). However, my C skills are too limited to correctly follow constructs with `goto`s into the middle of if statements, so this is my best attempt -- sorry.

### DIFF
--- a/server/op_editconfig.c
+++ b/server/op_editconfig.c
@@ -28,6 +28,9 @@
 #include "common.h"
 #include "operations.h"
 
+/* there is no enum for "no move", and the name SR_MOVE_LAST suggests that it might be the last item of an enum */
+#define SR_MOVE_INVALID_SENTINEL 666
+
 static enum NP2_EDIT_OP
 edit_get_op(struct lyd_node *node, enum NP2_EDIT_OP parentop, enum NP2_EDIT_DEFOP defop)
 {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi Jan,
I have no problems with it and believe it fixes the problem, merging. As for the current *netopeer2*, it does not perform the `edit-config` itself but rather leaves it up to *sysrepo*. And there are tests for moving user-ordered lists.

Regards,
Michal

jktjkt:
> As for the current _netopeer2_, it does not perform the `edit-config` itself but rather leaves it up to _sysrepo_. And there are tests for moving user-ordered lists.

Awesome, thanks for covering these.

PR #649, opened 2020-06-05
script.sh: take DESTDIR into account

### LINKED ISSUE #648: setup.sh script does not take DESTDIR into account for yang modules
The cmake `YANG_MODULE_DIR` variable is set to `"/usr/share/yang/modules/netopeer2`. The cmake installation will add the DESTDIR prefix to the destination for the yang modules. But the `setup.sh` script tries to install the yang modules with sysrepoctl from the `YANG_MODULE_DIR` without the DESTDIR prefix.

hthiery:
When installing yang modules add DESTDIR prefix to the NP2_MODULE_DIR.

Fixes #648

Signed-off-by: Heiko Thiery <heiko.thiery@gmail.com>

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,7 +8,7 @@ fi
 
 # avoid problems with sudo path
 SYSREPOCTL=`su -c "which sysrepoctl" $USER`
-MODDIR=${NP2_MODULE_DIR}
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
okay, no problem, thanks.

Regards,
Michal

PR #657, opened 2020-06-18
fixed several <error-tags> (issue#654)

### LINKED ISSUE #654: <error-tag>'operation-failed' instead of 'lock-denied'
Hi Michal et al,

when trying to `<lock>` an already locked datastore, netopeer2 replies correctly with an `<rpc-error>` message, but the `<error-tag>` element just contains '_operation-failed_' but shall be '_lock-denied_'.

When looking into the logs, netopeer2 seems to obtain the correct reason from sysrepo2

> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [ERR]: SR: Requested resource already locked

but then it maps that just to the tag 'operation-failed'. Could you look at this?

Thanks and regards,
Jens

Here a longer section of the logfile:
 
> Jun 15 15:21:09 58e962bf7514 netopeer2-server[3442]: Published event "rpc" "/ietf-netconf:lock" with ID 6 priority 0 for 1 subscribers.
> Jun 15 15:21:10 58e962bf7514 netopeer2-server: [INF]: SR: Published event "rpc" "/ietf-netconf:lock" with ID 6 priority 0 for 1 subscribers.
> Jun 15 15:21:10 58e962bf7514 netopeer2-server[3442]: Processing "/ietf-netconf:lock" "rpc" event with ID 6 priority 0 (remaining 1 subscribers).
> Jun 15 15:21:10 58e962bf7514 netopeer2-server: [INF]: SR: Processing "/ietf-netconf:lock" "rpc" event with ID 6 priority 0 (remaining 1 subscribers).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Module "yang" is locked by session 14 (NC SID 5).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Failed processing of "rpc" event with ID 6 priority 0 (remaining 1 subscribers).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [ERR]: SR: Module "yang" is locked by session 14 (NC SID 5).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [INF]: SR: Failed processing of "rpc" event with ID 6 priority 0 (remaining 1 subscribers).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Requested resource already locked
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [ERR]: SR: Requested resource already locked
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Event "rpc" with ID 6 priority 0 failed (Requested resource already locked).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [WRN]: SR: Event "rpc" with ID 6 priority 0 failed (Requested resource already locked).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Failed to send an RPC (User callback failed).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [ERR]: NP: Failed to send an RPC (User callback failed).
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Session 6: thread 0 event new RPC.
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [INF]: NP: Session 6: thread 0 event new RPC.
> Jun 15 15:21:12 58e962bf7514 netopeer2-server[3442]: Session 6: thread 0 event reply error.
> Jun 15 15:21:12 58e962bf7514 netopeer2-server: [INF]: NP: Session 6: thread 0 event reply error.

jkmuller:
Fixes #654 
This is an attempt to fix following  `<error-tag>` values which were reported as `operation-failed` but shall be more specific:
- `<lock-denied>`
- `<in-use>`
- `<data-exist>`

The proposed solution here is based on the initial fix 2bd857b - to parse the message parameter in np2srv_err_sr().

There might be many more error-tags missing, though, which needs to be added case-by-case. 

IMO, more SR_ERR_XXX values for each error-tag use case would be easier to handle, though.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -344,9 +344,28 @@ err_access_denied:
         /* fallthrough */
     default:
         if (strstr(message, "authorization failed")) {
+            // access-denied
             goto err_access_denied;
-        } else if (strstr(message, "is already locked")) {
+        } else if (strstr(message, "is already locked") ||
+                   strstr(message, "Module \"yang\" is locked by session")) {
+            // lock-denied
             goto err_lock_denied;
+        } else if (strstr(message, "is locked by session")) {
+            // in-use
+            e = nc_err(NC_ERR_IN_USE, NC_ERR_TYPE_PROT);
+            nc_err_set_msg(e, message, "en");
+            if (xpath) {
+                nc_err_set_path(e, xpath);
+            }
+            break;
+        } else if (strstr(message, "already exists")) {
+            // data-exists
+            e = nc_err(NC_ERR_DATA_EXISTS);
+            nc_err_set_msg(e, message, "en");
+            if (xpath) {
+                nc_err_set_path(e, xpath);
+            }
+            break;
         } else if (strstr(message, "Source and target")) {
             e = nc_err(NC_ERR_INVALID_VALUE, NC_ERR_TYPE_PROT);
             nc_err_set_msg(e, message, "en");


jkmuller:
Hello @michalvasko do you have an opinion about that? It's been open for 7 days now. Please see also related issues. Thanks, Jens

michalvasko:
Hi,
firstly, I know I usually react much sooner but I have been working on some features that had a high priority so I did not have the time to reply to issues.

Now, you are right that this is hardly an ideal solution but to actually get more `SR_ERR_*` values a *sysrepo* API change (the error structure) is required and we will wait with that for now. Also, the messages will always have to be examined because there is no 1:1 mapping of `SR_ERR` values and NETCONF error tags.

Having said that, I think your changes are fine but wanted to briefly test them first. Please wait a few more days.

Regards,
Michal

michalvasko:
Hi,
I have just tried it and you have reversed the fix I did before, I am again getting
```
ERROR
        type:     application
        tag:      operation-failed
        severity: error
        message:  Module "yang" is already locked by this session 13 (NC SID 1).
```
but with the current `devel` I got
```
ERROR
        type:     protocol
        tag:      lock-denied
        severity: error
        message:  Module "yang" is already locked by this session 10 (NC SID 1).
        SID:      1
```

Regards,
Michal

jkmuller:
Ok, I will check this. What about the other missing error codes?

michalvasko:
Hi,
you are free to check all of them and provide a patch or at least report them. It is not a critical problem and I have more important work currently.

Regards,
Michal

jkmuller:
Sorry for the delay on this.
I have checked the issue with the initial PR changes and it seems your and my test case trigger different `sr_errinfo()` calls in sysrepo for "already locked / in-use" case.
1. https://github.com/sysrepo/sysrepo/blob/22c0226b009f9190c7e2c8853c0c2d7616d1aaeb/src/shm_mod.c#L79
                `sr_errinfo_new(&err_info, SR_ERR_LOCKED, NULL, "Module \"%s\" is %s by session %u (NC SID %u).",`
2. https://github.com/sysrepo/sysrepo/blob/7920e6a336cf9d266cac17d5b459f0d9d42af754/src/sysrepo.c#L2735
            `sr_errinfo_new(&err_info, SR_ERR_LOCKED, NULL, "Module \"%s\" is already locked by this session %u (NC SID %u).",`
Thus I broke your case #2. This is hopefully included now in the updated PR. Nevertheless I would consider that all as workaround until the sysrepo error API can return NC error-tags as well, as the string based determination is pretty error-prone...
Thanks,
Jens

michalvasko:
Hi,
okay then. Like I said, it can be fixed only with a major SO version increase, which is always a pain to do. Also, may I ask what you need all these NETCONF-standardized errors for?

Regards,
Michal

jkmuller:
We run a kind of an NC conformance test against the NC stack. These test steps expect NC reply/rpc-error/error-tag according to RFC6241...
Thanks
Jens

michalvasko:
Thanks, just what I thought.

PR #666, opened 2020-06-30
commands: switch to X509_get0_notAfter

neheb:
Normal get function is deprecated as of OpenSSL 1.1.


michalvasko:
Hi,
okay, no problem, but we always merge only into `devel` branch.

Regards,
Michal

PR #707, opened 2020-08-25
scripts FEATURE improve syntax

jsmolic:


michalvasko:
Hi,
firstly, please target `devel` branch, there were some improvements recently. Secondly, did you try these scripts at least with `bash`, `dash` and `ash`? I do not know what other common default shells there are.

jsmolic:
> Hi,
> firstly, please target `devel` branch, there were some improvements recently. Secondly, did you try these scripts at least with `bash`, `dash` and `ash`? I do not know what other common default shells there are.

`merge_config` and `merge_hostkey` scripts work well on `ash`, `bash`, `dash,` and `zsh`. I'm having some issues with `setup.sh` on `dash` and `zsh`, but regardless of my changes that one did not work for me with `zsh` or `dash`, so I'll try to fix that up

michalvasko:
Okay, since you changed the interpreter in each of the scripts to the default one, I want to merge it only after the scripts works with all the common shells. Thanks for changing it that way. Also, please fix the conflicts you now have with the `devel`. Like I said, there were some improvements made.

jsmolic:
@michalvasko Sure, we should fix it so that it works on all shells.\
I resolved all conflicts and pushed again.

`eval` is not working for me in the script when using `dash`, while on `zsh` this line causes problems: https://github.com/CESNET/netopeer2/blob/5aa5e1953022f90c59f8493e60cdcf8cf8d0f04e/scripts/setup.sh#L117 (`condition expected: <`)

michalvasko:
Do you mean that you cannot fix it and I should take a look?

jsmolic:
> Do you mean that you cannot fix it and I should take a look?

No need, will look into it

michalvasko:
Okay.

jsmolic:
I got it to work on `zsh` only. `setup.sh` uses some features that are not POSIX compliant, so it doens't run properly on `dash`. I might tackle that when I catch more free time. I left the scripts compatible only with bash for now.

I ran the scripts through ShellCheck and made some general syntax improvements which could be useful, what do you think?

michalvasko:
Hi,
okay, thanks. So I guess it could be merged because the compatibility is the same.

Just one possible improvement, I have put all the features to enable with `-e` in the module list because it then could be used directly when installing the modules. However, it seems that is no longer the case and they server no purpose anymore so those `-e` should probably be removed and the regexps modified.

michalvasko:
Um, I meant just removing the `-e` parameters, the features themselves must stay to be enabled.

jsmolic:
> Um, I meant just removing the `-e` parameters, the features themselves must stay to be enabled.

Right :sweat_smile: I think I got it now

PR #708, opened 2020-08-26
Ignore key reading failure if not running as root

syyyr:
Running Netopeer2 as a normal user is a good use-case for testing.

In my test, I try to get all data available (with a `/*` filter). through sysrepo, but this callback errored out, when root access wasn't available.

cc @jktjkt

### DIFF
--- a/src/netconf_server_ssh.c
+++ b/src/netconf_server_ssh.c
@@ -431,11 +431,16 @@ np2srv_endpt_ssh_auth_users_oper_cb(sr_session_ctx_t *UNUSED(session), const cha
         }
         f = fopen(path, "r");
         if (!f) {
-            if (errno != ENOENT) {
+            if (errno != ENOENT && errno != EACCES) {
                 ERR("Opening \"%s\" authorized key file failed (%s).", path, strerror(errno));
                 free(path);
                 goto cleanup;
             }
+
+            if (errno == EACCES) {
+                VRB("Skipping \"%s\" authorized key file (%s).", path, strerror(errno));
+            }
+
             free(path);
             continue;
         }


michalvasko:
Hi,
I am not really sure about this. In effect, you are skipping authorized SSH keys that you have no access to. You will then advertise this incorrect information in the `operational` datastore. While it may be a good use-case for testing, *netopeer2-server* is meant to be run as root because otherwise you get similar problems resulting in degraded functionality.

jktjkt:
Hi @michalvasko , we're using Netopeer2 in our integration tests. It acts as a NETCONF server to verify that our NETCONF client works correctly. Please note that this patch only changes the behavior when the user is:
- running the NETCONF server as an unprivileged user,
- the Unix permissions prevent that user from reading the keys.

I think that this is pretty much the exact same approach that [NACM is using](https://tools.ietf.org/html/rfc8341#section-3.2.4) when it says:

> Data nodes to which the client does not have read access are silently omitted, along with any descendants, from the `<rpc-reply>` message.

Without this patch, we would have to build as root, or come up with elaborative methods on faking root. In previous version of sysrepo, we had to do this with a non-default build of `fakeroot`, and it was a pain in the rear to maintain.

Feel free to scope this change with a CMake option if you have to. Also, perhaps it would be better to only do this silent skip on `EACCESS`?

michalvasko:
Hi,
okay, I guess, although I am not sure about the comparison to NACM. Anyway, yes, change it so that only `EACCESS` is ignored. Also, the UID does not have to be checked but instead print an INF message informing about an authorized key being skipped. Thanks.

syyyr:
Okay, I made the changes and rebased

syyyr:
Okay, I made the changes.

src/netconf_server_ssh.c
```diff
@@ -431,11 +431,16 @@ np2srv_endpt_ssh_auth_users_oper_cb(sr_session_ctx_t *UNUSED(session), const cha
...
         if (!f) {
-            if (errno != ENOENT) {
+            if (errno != ENOENT && errno != EACCES) {
                 ERR("Opening \"%s\" authorized key file failed (%s).", path, strerror(errno));
                 free(path);
                 goto cleanup;
             }
+
+            if (errno == EACCES) {
+                VRB("Skipping \"%s\" authorized key file. (%s)", path, strerror(errno));
```
michalvasko, line 441: Sorry for bothering, but one last detail, please put `.` after the parentheses with the error to follow the message above and all the other, thanks.

michalvasko (changes requested):

src/netconf_server_ssh.c
syyyr: Sure, no problem

jktjkt (approved):

michalvasko (approved):

syyyr:
Thanks for merging this!

PR #710, opened 2020-09-01
build FEATURE add env var sysrepo tools override

syyyr:
In my dev environment, there is no sysrepoctl or sysrepocfg in `PATH`. I usually have two versions of the netconf stack installed (at the same time): one with sanitizers and one without, so the scripts wouldn't know which binary to use (and they are usually incompatible). I added an env var override for sysrepoctl and sysrepocfg.

Also, feel free to (completely) change the commit message if you want, I'm not sure if I got that right.

Thanks

### DIFF
--- a/scripts/merge_config.sh
+++ b/scripts/merge_config.sh
@@ -2,8 +2,11 @@
 
 set -e
 
+# optional env variable override
+if [ -n "$SYSREPOCFG_EXECUTABLE" ]; then
+    SYSREPOCFG="$SYSREPOCFG_EXECUTABLE"
 # avoid problems with sudo PATH
-if [ `id -u` -eq 0 ]; then
+elif [ `id -u` -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


syyyr:
I messed up the conditions a little bit, but it should work now.

syyyr:
Okay, NOW it should be okay, apparently I don't know how scripting works :100:

michalvasko:
Hi,
okay, I suppose (although there sure are a lot env vars in those scripts), but please at least mention these variables at the script beginning, as being optional.

syyyr:
I'm not sure I understood you correctly, but I changed the comments to say "optional"

michalvasko:
Right, the variables are mentioned only in `setup.sh`, not in the other scripts. So fine, whatever.

syyyr:
Okay, thanks.

PR #731, opened 2020-10-07
Fix build with older cmake versions

rjarry:
Fix the following error with cmake 2.8.12:

```
 ...
 -- Found LibSSH: /usr/include (found suitable version "0.7.6", minimum required is "0.7.1")
 ...
 CMake Error at cli/CMakeLists.txt:29 (message):
  libnetconf2 supports SSH but libssh was not found, CLI compilation failed!
```

There was a mixup between `LibSSH_FOUND` and `LIBSSH_FOUND`. The former seems not to be propagated to cmake subdirectories with older cmake versions (2.8.12 for example). Same for `OpenSSL_FOUND`.

```
 CMakeModules/FindLibSSH.cmake:#  LIBSSH_FOUND - system has LibSSH
 /usr/share/cmake/Modules/FindOpenSSL.cmake:#  OPENSSL_FOUND - system has the OpenSSL library
```

Use upper case variable names which are working with all versions of cmake.

Fixes: 1628962f725f ("build BUGFIX non-standard paths of lnc2 lib/headers")


michalvasko:
Hi Robin,
`cmake` incompatibilities are a nightmare, thanks for the patch. Merging.

rjarry:
Thanks Michal :+1:

rjarry:
Maybe switch to [meson](https://mesonbuild.com/) someday :)

michalvasko:
First time I hear about *meson* so I took a quick look. It seems fairly usable although, for example, dependencies seem to me are quite messy having many "methods" of searching for them (including `cmake` modules). So in case there is a problem, you first have to figure out what method was actually used. But thanks for mentioning it anyway.

rjarry:
Yes, dependency searching is quite ugly with any build automation tool I know...

PR #732, opened 2020-10-09
scripts: improve error handling when Bash is not at /bin/bash

jktjkt:
Two commits here:

## scripts: do not rely on /bin/bash
    
I'm playing around with NixOS which has bash, but not under `/bin/bash`. This has caused `scripts/setup.sh` which gets run during `make install` to fail silently.
    
Personally, I think that not having /bin/bash is silly, but hey, and other scripts, including ones in sysrepo, already use this construct with `/usr/bin/env`, so let's follow their lead.

## build: fail `make install` if scripts fail
    
Failures of these shell scripts were not visible. In case they managed to output something, that something was shown, but if the script invocation itself failed, the user did not see any warning. It was nicely visible on NixOS which doesn't have /bin/bash at a standard `/bin/bash`, but instead on some strange forsaken place.


michalvasko:
Hi,
as for the return values, it makes sense and I have nothing against it. However, it seems to me that `env(1)` is a GNU tool and there can be no strict dependency on such binaries.

jktjkt:
`env` is [part of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html#tag_20_39), whereas the same document also [explicitly says that there's no guarantee that the plain old `sh` lives in `/bin/sh`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#tag_20_117_16) (wow!). I propose to merge this because sysrepo is already using the `/usr/bin/env sh`, and because it will make my life easier.

michalvasko:
Okay, I just was not sure because my manual page was from GNU. Merging.

PR #764, opened 2020-11-19
Add option for setting PID

syyyr:
Hi, now that there options for separating the sysrepo repository and shm prefix via environmental variables, the only thing that disallows launching more netopeer2 processes is the pidfile. This PR adds on option to set a custom pid file.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1042,10 +1042,11 @@ print_version(void)
 static void
 print_usage(char* progname)
 {
-    fprintf(stdout, "Usage: %s [-dhV] [-U (path)] [-m mode] [-u uid] [-g gid] [-v level] [-c category]\n", progname);
+    fprintf(stdout, "Usage: %s [-dhV] [-p pid] [-U (path)] [-m mode] [-u uid] [-g gid] [-v level] [-c category]\n", progname);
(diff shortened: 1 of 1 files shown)


jktjkt (approved):
And just to clarify, the reason for running several `netopeer2-server` processes in the first place is automated unit tests of NETCONF-talking software -- the [`netconf-cli`](https://github.com/CESNET/netconf-cli) in our case. It is very convenient just to launch `netopeer2-server` and react to sysrepo callbacks from the test suite.

src/main.c
```diff
@@ -1076,6 +1077,7 @@ main(int argc, char *argv[])
     int c, *idx, i;
     int daemonize = 1, verb = 0;
     int pidfd;
+    const char* pidfile = NP2SRV_PID_FILE_PATH;
```
michalvasko, line 1080: Minor formatting problem, use `char *` please.

michalvasko (changes requested):
Generally seems fine, thanks.

src/main.c
syyyr: done

michalvasko (approved):

PR #802, opened 2021-01-11
scripts: Limit activity to the startup datastore

jktjkt:
This script used to check the `running` datastore for presence of configuration statements. That doesn't work when Netopeer2 is not running. Make sure that we only touch the `startup` DS, and delegate any updates of the `running` DS to the usual mechanism (i.e., upon the first subscription).

### DIFF
--- a/scripts/merge_config.sh
+++ b/scripts/merge_config.sh
@@ -14,7 +14,7 @@ fi
 KS_KEY_NAME=genkey
 
 # check that there is no listen/Call Home configuration yet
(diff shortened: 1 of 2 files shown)


jktjkt:
Hmm, this doesn't work in all cases, either. I **think** that checking the content of the `startup` DS instead of the running one is better, but the `copy-config` at the end is apparently still required. Strange, I would have expected that `running` gets a copy of `startup` only upon the first subscription (i.e., when `netopeer2-server` starts)...

michalvasko:
Look at [the 3rd paragraph](https://netopeer.liberouter.org/doc/sysrepo/devel/html/index.html#datastores) for explanation. That being said, it may make more sense to check the startup configuration for any `ietf-netconf-server` data, I am not sure. But the idea simply was that whenever something was configured (if only into `running` datstore with `startup` being empty), it should not be overwritten. Hence `running` datastore is checked.

PR #806, opened 2021-01-18
Shorten UNKNWON to UNK to avoid buffer overflow

gwendlan:
Print src and fmt in case of memory shortage to have at least something (__FILE__,__LINE__ always prints log.c:49)

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -46,7 +46,7 @@ np2log(int priority, const char *src, const char *fmt, ...)
     if (np2_stderr_log) {
         format = malloc(11 + strlen(fmt) + 2);
         if (!format) {
-            fprintf(stderr, "[ERR]: Memory allocation failed (%s:%d)", __FILE__, __LINE__);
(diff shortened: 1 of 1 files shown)


michalvasko:
Okay, shortening the flag makes sense but I do not know what that additional information in case of memory allocation error is for. More importantly, you should never get an `UNKNOWN` message, so that is a bug. Can you provide more information about that?

gwendlan:
The "UNKNOWN" case should never occur, I agree. Still I would shorten it to three letters.

The intention of the changed print in case of memory shortage is, if memory allocation fails, then all one sees in the logfile is "out of memory in log.c:49". No clue as to where the log function was called from. By providing src (a two letter code of functional area) and the original format string one is given at least a tiny clue to guess where it the log function was called from.

michalvasko:
Okay then, I suppose, but I still doubt this PR will make any practical difference.

Issue #811, opened 2021-01-22
RFC: user management -- passwords, SSH pubkeys, etc


jktjkt:
Hi, do you have plans for supporting setting of user passwords and "systemwide" SSH public keys for authentication? We will need something like that within a month or so, and I noticed that a basic read-only data retrieval is already there via [`np2srv_endpt_ssh_auth_users_oper_cb`](https://github.com/CESNET/netopeer2/blob/0d2e6e516c4f0962b86f5da2a26334dbd82c1934/src/netconf_server_ssh.c#L308).

We might have to handle that via our "misc stuff for YANG" repo, and it might need disabling this particular CB, so just wondering about your plans and (rough) roadmaps here.

michalvasko:
I am not sure what exactly you are asking so I am just explain how it works now. Since all the YANG modules are still drafts, *netopeer2* reads whatever information it can form the system despite some of it being defined also in some of the YANG modules. This system information includes SSH usernames for authentication (system users) or authorized keys for each of these users (read from user-specific file `~/.ssh/authorized_keys`).

The plan is to have all the authentication information strictly NETCONF-specific and configured in the corresponding YANG modules but for that the modules must first be published as RFCs. So there is no plan for this at all for now.

jktjkt:
Thanks. As you know, we're building "appliances", and it's possible to access these not only via NETCONF, but also directly over SSH, where the user gets dumped into a [user-friendly CLI that talks to the appliance's sysrepo natively -- that's the `sysrepo-cli`](https://github.com/CESNET/netconf-cli) by @syyyr. We have to manage access to these entry points, somehow, and it makes sense to use the same authentication database for NETCONF and this CLI-via-SSH. So as a minimal feature, we need a way of changing the systemwide (PAM) password of a user, and to be able to create/read/modify/delete the users' SSH pubkeys. I see that this can be done via, e.g., the `ietf-system` module (released a long time ago), or `ietf-netconf-server` + `ietf-ssh-server` combination (still in drafts), or a custom module. I also see that Netopeer2 already has some code to populate the `operational` DS of the draft models with user info.

Are you going to (eventually, some time after these RFCs materialize) change Netopeer2 to only use the YANG data with no relation to the users' SSH keys, `getpwnam`, etc?

michalvasko:
> Are you going to (eventually, some time after these RFCs materialize) change Netopeer2 to only use the YANG data with no relation to the users' SSH keys, getpwnam, etc?

[...]

PR #812, opened 2021-01-22
Compile-time options for specifying custom paths of authorized_keys

jktjkt:
Our appliance uses a read-only rootfs, and we're therefore managing the users' SSH pubkeys within a config-only partition. That's outside of their individual home directories.

Cc: #811

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,12 @@ if(NOT PIDFILE_PREFIX)
     set(PIDFILE_PREFIX "/var/run")
 endif()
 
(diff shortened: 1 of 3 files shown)


michalvasko:
Okay, it seems you need something like this so it can be merged, but the options are quite wild. But please, at least add a `cmake` check that the pattern really contains one `%s`, avoids SEGFAULTs.

jktjkt:
> but the options are quite wild

Yeah, I just preferred not to code my own str-replace in C :)

> But please, at least add a `cmake` check that the pattern really contains one `%s`, avoids SEGFAULTs.

Done.

michalvasko:
Okay, merging.

Issue #818, opened 2021-02-01
segmentation fault when operational data callback takes more time that rpc timeout


rjarry:
The default sysrepo timeouts are:

```c
/** default timeout for operational subscription callback (ms) */
#define SR_OPER_CB_TIMEOUT 5000
/** default timeout for RPC/action subscription callback (ms) */
#define SR_RPC_CB_TIMEOUT 2000
```

Since commits https://github.com/CESNET/netopeer2/commit/d1e0217a606e8719f6479b340479529b391740ad and https://github.com/CESNET/netopeer2/commit/7c5b112235449ba7ae7d6da6bdbbf0746d1f5da0, `netopeer2-server` uses the default sysrepo timeouts unless the `-t` argument is specified.

When a client connects and sends a `<get>` RPC, the call graph is as follows:

```
MAIN THREAD

np2srv_rpc_cb  "<get>"
  |- sr_rpc_send_tree                    (timeout=2s)
  |   `- sysrepo code which notifies a second thread
  `- returns error to client, client closes session

SECOND THREAD

sr_process_events
 `- np2srv_rpc_get_cb
     |- np2srv_get_rpc_data              (timeout=5s)
     |- np_get_nc_sess_user              (session has been closed by client)
     `- ncac_check_data_read_filter      (with user=NULL)
         `- getpwnam_r(NULL)             segmentation fault
```

How to reproduce:

```console
foo@bar:~$ docker run -it --name sysrepo --rm sysrepo/sysrepo-netopeer2:latest bash -li
...
Digest: sha256:266d2cdfe2c34075c8921b98303322f51e3a7348d8776783f99531d5a3aa0563
Status: Downloaded newer image for sysrepo/sysrepo-netopeer2:latest
root@e450e566bc54:/# gcc application.c -Wall -Wextra -Werror -o application -lyang -lsysrepo
root@e450e566bc54:/# gcc client.c -Wall -Wextra -Werror -o client -lyang -lnetconf2
root@e450e566bc54:/# sysrepoctl -i example.yang
root@e450e566bc54:/# ulimit -c unlimited
root@e450e566bc54:/# ./application &
[1] 88
root@e450e566bc54:/# netopeer2-server -U -d -v 1
[2] 181
root@e450e566bc54:/# ./client 
[WRN]: NP: Session 1 uses a transport protocol not supported by ietf-netconf-monitoring, will not be monitored.
[...]

PR #819, opened 2021-02-01
acm: fix segmentation fault when session is disconnected

rjarry:
If NETCONF session was disconnected while waiting for sr_get_data, before filtering the returned data np_get_nc_sess_user() returns NULL, then user will be NULL as well.

This translates in ncac_getpwnam calling getpwdnam_r() with user=NULL. getpwdnam_r() may call _nss_compat_getpwnam_r() which accesses user[0] without checking if user is NULL first.

This causes a segmentation fault:

```
 Stack trace:
 #0      _nss_compat_getpwnam_r (libnss_compat.so.2)
 #1      getpwnam_r (libc.so.6)
 #2      n/a (netopeer2-server)
 #3      n/a (netopeer2-server)
 #4      ncac_check_data_read_filter (netopeer2-server)
 #5      np2srv_rpc_get_cb (netopeer2-server)
 #6      n/a (libsysrepo.so.5)
 #7      n/a (libsysrepo.so.5)
 #8      sr_process_events (libsysrepo.so.5)
 #9      n/a (libsysrepo.so.5)
 #10     start_thread (libpthread.so.0)
 #11     __clone (libc.so.6)
```

Avoid calling ncac_check_data_read_filter if np_get_nc_sess_user returns NULL.

Also, for good measure, check for the user argument value in ncac_getpwnam and avoid calling getpwdnam_r if it is NULL.

Fixes: #818
Link: https://sourceware.org/git/?p=glibc.git;a=blob;f=nss/nss_compat/compat-pwd.c;h=f536754559ad;hb=HEAD#l819

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -76,6 +76,7 @@ np2srv_ntf_new_cb(sr_session_ctx_t *UNUSED(session), const sr_ev_notif_type_t no
     struct nc_server_notif *nc_ntf = NULL;
     struct nc_session *ncs = (struct nc_session *)private_data;
     struct lyd_node *ly_ntf = NULL;
+    const char *username;
     NC_MSG_TYPE msg_type;
     char buf[26], *datetime;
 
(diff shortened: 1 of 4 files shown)


michalvasko:
Thanks for the patch but I decided to be more restrictive and if the username is retrieved immediately, this specific use-case should not happen (NETCONF session being closed when the username is retrieved). So any other such cases, it is an internal error.

PR #833, opened 2021-02-11
netconf BUGFIX server crash protection

lePici:
Refs #818
The server will not crash due to username getter function.
However, the internal error still remains unresolved.

### DIFF
--- a/src/netconf.c
+++ b/src/netconf.c
@@ -193,6 +193,22 @@ np2srv_rpc_get_cb(sr_session_ctx_t *session, const char *op_path, const struct l
     int rc = SR_ERR_OK;
     struct ly_set *nodeset;
     sr_datastore_t ds = 0;
+    char *username = NULL;
+
+    /* Get username. It is assumed that right now the NETCONF session cannot end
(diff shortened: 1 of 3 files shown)


src/netconf.c
```diff
@@ -193,6 +193,13 @@ np2srv_rpc_get_cb(sr_session_ctx_t *session, const char *op_path, const struct l
     int rc = SR_ERR_OK;
     struct ly_set *nodeset;
     sr_datastore_t ds = 0;
+    const char *username;
+
+    /* Get username. It is assumed that right now the NETCONF session cannot end due to the lock. */
+    if (!(username = np_get_nc_sess_user(session))) {
```
michalvasko, line 199: The problem here is that you first get the username (which should be safe at this moment) but use it after too long time (including possibly getting operational data) when you can no longer guarantee that it was not freed. That is why you must duplicate it after you get it form the session, in all the uses except for `np2srv_diff_check_cb` because there the username is immediately used.

michalvasko (changes requested):

michalvasko (approved):

PR #865, opened 2021-03-11
Consideration of partial matching NACM rules corrected.

skolbel:
Precondition:
- A RADIUS server with the corresponding shared-secret has been configured.
- The following NACM rules have been configured:
```xml
    <nacm:nacm xmlns:nacm="urn:ietf:params:xml:ns:yang:ietf-netconf-acm" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0">
        <nacm:rule-list>
            <nacm:name>EXAMPLE</nacm:name>
            <nacm:group>EXAMPLE</nacm:group>
            <nacm:rule>
                <nacm:name>denyRadiusSecret</nacm:name>
                <nacm:action>deny</nacm:action>
                <nacm:path xmlns:sys="urn:ietf:params:xml:ns:yang:ietf-system">/sys:system/sys:radius/sys:server/sys:udp/sys:shared-secret</nacm:path>
            </nacm:rule>
            <nacm:rule>
                <nacm:name>denyRadiusSecret</nacm:name>
                <nacm:access-operations>read</nacm:access-operations>
                <nacm:action>permit</nacm:action>
                <nacm:path xmlns:sys="urn:ietf:params:xml:ns:yang:ietf-system">/sys:system/sys:radius/sys:server/*</nacm:path>
            </nacm:rule>
            <nacm:rule>
                <nacm:name>allowRead</nacm:name>
                <nacm:access-operations>*</nacm:access-operations>
                <nacm:action>permit</nacm:action>
                <nacm:module-name>*</nacm:module-name>
            </nacm:rule>
        </nacm:rule-list>
    </nacm:nacm>
```
Expectation:
The result of a get-config shall not contain the shared-secret.

Observation:
The shared-secret is contained in the RPC reply.

Cause:
This behavior is because the "allowRead" rule finally grants access and the variable partial_access is initialized
using the wrong value.
Additoinally, the PARTIAL_... conditions are not considered correctly within the function "ncac_allowed_node()".
So, a set of rules may result in both conditions PARTIAL_DENY and PARTIAL_PERMIT for one node.
Also the evaluation of partial_access should always take place.

Correction:
I have used bits for encoding the PARTIAL... conditions, changed the evaluation of partial_access at the
end of the function and moved this code behind the label "cleanup".

### DIFF
--- a/src/netconf_acm.c
+++ b/src/netconf_acm.c
@@ -1063,7 +1063,12 @@ ncac_allowed_node(const struct lyd_node *node, const char *user, uint8_t oper)
     struct ly_ctx *ly_ctx;
     char **groups, *path;
     uint32_t i, j, group_count;
-    enum ncac_access access = NCAC_ACCESS_DENY, partial_access = NCAC_ACCESS_DENY;
+    enum ncac_access access = NCAC_ACCESS_DENY;
+    enum {
+        RULE_PARTIAL_MATCH_NONE   = 0,
+        RULE_PARTIAL_MATCH_PERMIT = 1,
+        RULE_PARTIAL_MATCH_DENY   = 2
+    } partial_access = RULE_PARTIAL_MATCH_NONE;
     int path_match;
 
     assert(oper);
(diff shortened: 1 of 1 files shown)


michalvasko:
Okay, I think you are right and thanks for providing a fix.

PR #907, opened 2021-05-27
Tests libyang2

Brwo:
Create test for some basic RPCs.
Cancel test currently segfaults.

### DIFF
--- a/tests/test_rpc.c
+++ b/tests/test_rpc.c
@@ -33,72 +33,89 @@
 #include <nc_client.h>
 
 #include "np_test.h"
-#include "np_test_config.h"
+#include "np_test_config.h.in"
+
+#define LOCK_FAIL_TEMPLATE "<rpc-reply xmlns=\"urn:ietf:params:xml:ns:netconf:base:1.0\" message-id=\"%ld\">\n"\
+        "  <rpc-error>\n"\
+        "    <error-type>protocol</error-type>\n"\
+        "    <error-tag>lock-denied</error-tag>\n"\
+        "    <error-severity>error</error-severity>\n"\
+        "    <error-message lang=\"en\">Access to the requested lock is denied because the lock is currently held by another entity.</error-message>\n"\
+        "    <error-info>\n"\
+        "      <session-id>%d</session-id>\n"\
+        "    </error-info>\n"\
+        "  </rpc-error>\n"\
+        "</rpc-reply>\n"
 
 NP_GLOB_SETUP_FUNC
+
+static void
+test_types(void **state)
+{
+    struct nc_rpc *rpc;
+
+    (void)state;
+
+    /* Test that all constructors create the right type of rpc */
+    rpc = nc_rpc_lock(NC_DATASTORE_RUNNING);
+    assert_int_equal(NC_RPC_LOCK, nc_rpc_get_type(rpc));
+    nc_rpc_free(rpc);
+
+    rpc = nc_rpc_unlock(NC_DATASTORE_RUNNING);
+    assert_int_equal(NC_RPC_UNLOCK, nc_rpc_get_type(rpc));
+    nc_rpc_free(rpc);
+
+    rpc = nc_rpc_get("", NC_WD_ALL, NC_PARAMTYPE_CONST);
+    assert_int_equal(NC_RPC_GET, nc_rpc_get_type(rpc));
+    nc_rpc_free(rpc);
+
+    rpc = nc_rpc_kill(NC_DATASTORE_RUNNING);
+    assert_int_equal(NC_RPC_KILL, nc_rpc_get_type(rpc));
+    nc_rpc_free(rpc);
+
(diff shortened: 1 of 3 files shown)


tests/test_rpc.c
```diff
@@ -20,6 +20,10 @@
  * limitations under the License.
  */
 
+#include <libnetconf2/messages_client.h>
+#include <libnetconf2/netconf.h>
+#include <libyang/printer_data.h>
+#include <libyang/tree_data.h>
```
michalvasko, line 26: Never put headers before feature macros (`define X`), it can lead to some unexpected problems (when the headers you are including are actually recursively including the system header you defined the feature macro for so they are compiled without the macro defined and you get errors even though you have defined the feature macro).

tests/test_rpc.c
```diff
@@ -65,20 +131,25 @@ test_lock(void **state)
...
+    /* recieve reply, should yield error */
+    msgtype = nc_recv_reply(st->nc_sess2, rpc, msgid, 2000, &envp, &op);
     assert_int_equal(msgtype, NC_MSG_REPLY);
     assert_null(op);
+    assert_string_equal(LYD_NAME(lyd_child(envp)), "rpc-error");
     assert_int_equal(LY_SUCCESS, lyd_print_mem(&str, envp, LYD_XML, LYD_PRINT_SHRINK));
-    lyd_free_tree(envp);
+
+    char expected[500];
+    sprintf(expected, LOCK_FAIL_TEMPLATE, msgid, nc_session_get_id(st->nc_sess));
```
michalvasko, line 149: Unsafe and no reason for it, no problem with allocating the string dynamically (a few lines below). Also, variable declarations can only be at block start (in this case function start) but you can reuse `str2` or rename it to `expected`.

tests/test_rpc.c
```diff
@@ -96,32 +167,212 @@ test_lock(void **state)
...
     const struct CMUnitTest tests[] = {
-        cmocka_unit_test(test_lock),
+        cmocka_unit_test(test_nc_rpc_types),
+        cmocka_unit_test(test_rpc_lock),
+        cmocka_unit_test(test_rpc_unlock),
+        cmocka_unit_test(test_rpc_get),
+        cmocka_unit_test(test_rpc_kill),
+        cmocka_unit_test(test_rpc_commit),
+        cmocka_unit_test(test_rpc_discard),
+        cmocka_unit_test(test_rpc_cancel),
```
michalvasko, line 375: No need to call every test `*_rpc_*`, the whole test file is called `test_rpc` so it is implied.

michalvasko (changes requested):

tests/test_rpc.c
```diff
@@ -28,100 +28,325 @@
 #include <stdlib.h>
 #include <stdarg.h>
 
+#include <libnetconf2/messages_client.h>
+#include <libnetconf2/netconf.h>
+#include <libyang/printer_data.h>
+#include <libyang/tree_data.h>
```
michalvasko, line 34: All includes redundant, all `libnetconf2` client headers are in `nc_client.h` and all `libyang` headers in `libyang/libyang.h`.

tests/test_rpc.c
```diff
@@ -28,100 +28,325 @@
...
 
     nc_rpc_free(rpc);
+    lyd_free_tree(envp);
 
     /* unlock RPC */
     rpc = nc_rpc_unlock(NC_DATASTORE_RUNNING);
-    assert_non_null(rpc);
+    assert_int_equal(NC_MSG_RPC, nc_send_rpc(st->nc_sess, rpc, 1000, &msgid));
+    msgtype = nc_recv_reply(st->nc_sess, rpc, msgid, 2000, &envp, &op);
+    assert_int_equal(msgtype, NC_MSG_REPLY);
```
michalvasko, line 160: You should probably stick to either the assert, expected return, and function call on one line or put the assert on a separate line, not mix it. I prefer having the assert on separate line because otherwise there is a bit too much clutter.

tests/test_rpc.c
```diff
@@ -28,100 +28,325 @@
...
+    assert_string_equal(LYD_NAME(lyd_child(envp)), "ok");
 
-    /* send request */
-    msgtype = nc_send_rpc(st->nc_sess, rpc, 1000, &msgid);
-    assert_int_equal(msgtype, NC_MSG_RPC);
+    nc_rpc_free(rpc);
+    lyd_free_tree(envp);
+    /* TODO: check if lock prevents changes */
+}
+
```
michalvasko, line 168: Extra line.

tests/test_rpc.c
```diff
@@ -28,100 +28,325 @@
...
+    assert_int_equal(msgtype, NC_MSG_REPLY);
+    assert_null(op);
     assert_string_equal(LYD_NAME(lyd_child(envp)), "ok");
+
+    nc_rpc_free(rpc);
+    lyd_free_tree(envp);
+}
+
+static void
+test_get(void **state) {
```
michalvasko, line 230: `{` should be on a separate line, for all the functions.

michalvasko (changes requested):

PR #1021, opened 2021-09-24
misc Ci fixes

jktjkt:
## build: do not disable testing on release builds
    
There's no need to skip testing for release builds. On the contrary, given that the release build is something that the end users will run, it makes sense to test the release builds as well.

## docs: point to GitHub Actions for CI results


michalvasko:
Right, no problem, thanks.

Issue #1020, opened 2021-09-24
Parallel testing doesn't work


jktjkt:
I'm on the latest `devel` branch, and when I try running the tests in parallel, they fail. I see that the code sets up a per-test-instance sysrepo repository, so it looks like this is something that was supposed to work. Perhaps there's a conflict on a listening port and/or a shared instance of the server?

michalvasko:
They were not meant to work in parallel, all the setup stuff was intended to avoid failing tests because of a previous failed test. But since these tests take a lot of time it makes sense to support concurrent execution so it should now work.

jktjkt:
Thanks, this has helped, but I'm still getting test failures as of commit 7e80c5c. This is on a 16-core CPU, and I'm getting similar results as long as I enable *any* parallelization, even with `-j2`:
```
Test project /home/jkt/work/prog/_build/czechlight-gcc-11-ly2/Netopeer2
      Start  5: test_subscribe_filter
      Start  7: test_parallel_sessions
 1/16 Test  #7: test_parallel_sessions ...........***Failed   11.49 sec
[==========] Running 1 test(s).
[  FAILED  ] GROUP SETUP
[  ERROR   ] tests
-U/home/jkt/work/prog/_build/czechlight-gcc-11-ly2/Netopeer2/tests/repositories/test_parallel_sessions/np2.sock
Setup fail in /home/jkt/work/cesnet/gerrit/github/CESNET/Netopeer2/tests/np_test.c:78.
Setup fail in /home/jkt/work/cesnet/gerrit/github/CESNET/Netopeer2/tests/np_test.c:231.
[==========] 0 test(s) run.
[  PASSED  ] 0 test(s).

      Start 12: test_sub_ntf_advanced
 2/16 Test  #5: test_subscribe_filter ............Subprocess aborted***Exception:  11.71 sec
[==========] Running 14 test(s).
[  FAILED  ] GROUP SETUP
[  ERROR   ] tests
Test failed with exception: Segmentation fault(11)

      Start 15: test_yang_push_advanced
 3/16 Test #12: test_sub_ntf_advanced ............Subprocess aborted***Exception:  11.60 sec
[==========] Running 18 test(s).
[  FAILED  ] GROUP SETUP
[  ERROR   ] tests
Test failed with exception: Segmentation fault(11)

      Start 14: test_yang_push
 4/16 Test #15: test_yang_push_advanced ..........Subprocess aborted***Exception:  11.60 sec
[...]

PR #1029, opened 2021-09-30
tests: use relative paths for Unix sockets

jktjkt:
Linux limits the socket paths to about 108 characters, and my setup uses rather long prefixes for the build dir. Since a relative path works here just fine, use that.

It is possible to set the CWD of a test run via CMake's `WORKING_DIRECTORY` property, but the default works just fine here, so there's no need to tweak that. There's a potential regression if a developer runs the test executable directly, but doing that bypasses a lot of CMake/CTest features already, such as fixtures or environment variables which are already used by this project, so I think this is a safe fix.

Fixes: https://github.com/CESNET/netopeer2/issues/1020

### DIFF
--- a/tests/np_test.c
+++ b/tests/np_test.c
@@ -178,13 +178,13 @@ np_glob_setup_np2(void **state, const char *test_name)
     }
 
     /* generate path to socket */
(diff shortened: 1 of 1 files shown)


michalvasko:
Thanks, but I want to preserve the option to run the tests directly as I use it often. Should work now with relative paths in any case.

PR #1036, opened 2021-10-08
Better OOM handling (asprintf) and other fixes

jktjkt:
## Fix asprintf usage
    
This function doesn't touch the output string buffer pointer upon failure. Instead, -1 is returned.

This was flagged by `-Wunused-result`.

Fixes: 65dc00b
Cc: @tadeas-vintrlik 

## cmake: Allow overriding per-build-type CFLAGS
    
Unfortunately, CMake's story for providing *default* flags for debug vs. release builds is not great. There's the `CMAKE_C_FLAGS_<type>`, but these got appended instead of prepended to user's `CFLAGS`, and as such they override whatever the user has requested. Not good.

### DIFF
--- a/cli/commands.c
+++ b/cli/commands.c
@@ -1615,7 +1615,11 @@ cmd_knownhosts(const char *arg, char **UNUSED(tmp_config_file))
         }
         free(text);
 
-        ftruncate(fileno(file), written);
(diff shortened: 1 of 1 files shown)


jktjkt:
Both fixed, sorry.

PR #1085, opened 2021-11-29
Fix build when not using CMake's make generator

jktjkt:
CMake is a generator which can be used with other tools than just `make` -- `ninja` is a good candidate which tends to build a little faster. It's therefore a bug to try invoking `make` on some target directly.

See-also: https://github.com/sysrepo/sysrepo/pull/2647
Fixes: e70b9c9 tests UPDATE cleanup SHM files after tests

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -52,10 +52,14 @@ foreach(test_name IN LISTS tests)
     set_property(TARGET ${test_name} PROPERTY RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 endforeach(test_name)
 
+set(TEST_KILL_SERVER_COMMAND ${CMAKE_CURRENT_BINARY_DIR}/scripts/kill_np_server.sh)
+set(TEST_CLEAR_STATE_COMMAND rm -rf /dev/shm/_tests_np_*)
 if(${CMAKE_VERSION} VERSION_GREATER "3.7")
-    # tests cleanup fixture, keep repos with server log files
-    add_test(NAME tests_done COMMAND make test_clean_keep_repos)
-    set_tests_properties(tests_done PROPERTIES FIXTURES_CLEANUP tests_cleanup)
+    # tests cleanup fixtures, keep repos with server log files
+    add_test(NAME tests_kill_server COMMAND ${TEST_KILL_SERVER_COMMAND})
+    add_test(NAME tests_clear_state COMMAND ${TEST_CLEAR_STATE_COMMAND})
+    set_tests_properties(tests_kill_server PROPERTIES FIXTURES_CLEANUP tests_cleanup)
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -82,8 +83,7 @@ endif()
 
 # phony target for clearing all sysrepo test data
 add_custom_target(test_clean_keep_repos
-    COMMAND ${CMAKE_CURRENT_BINARY_DIR}/scripts/kill_np_server.sh
-    COMMAND rm -rf /dev/shm/_tests_np_*
+    COMMAND ${TEST_CLEAR_STATE_COMMAND}
```
michalvasko, line 86: This target is not even needed now that a variable is used, it was defined instead of that, so you can remove it.

michalvasko (changes requested):
Ideally, the shell script should be rewritten into a cmake skript (function or something), it is not needed as a standalone script (not originally my design).

jktjkt:
> Ideally, the shell script should be rewritten into a cmake skript (function or something), it is not needed as a standalone script (not originally my design).

In CMake, fixtures are required to be a `test`, not just an arbitrary `target`. Tests take only one `COMMAND`. As a result, I cannot easily do killing of processes *and* removal of files from CMake, at least not without a major refactoring (or, as an absolute minimum, requiring two separate tests, which is a little bit meh if you ask me). That's why I put it to the shell script.

The referenced commit is a regression which breaks the CI for us, so I need to get this fixed now. If you want to rearchitect this to a CMake script for some reason, more power to you, and I'll be happy to review your changes, but I cannot spend time implementing that today I'm afraid. Also, there's no CMake-portable way of killing processes, so you will still need to shell out to execute that `kill`, so...

tests/CMakeLists.txt
jktjkt: I don't understand this comment; do you want me to remove this target completely?

michalvasko: Yes.

michalvasko:
Right, so for now I would not mind having 2 tests in the fixtures (like you said, there is no way around it). One would run the script, another would clean the SHM.

Also, my general problem is with the scripts themselves, not having to run shell commands (that is only what `COMMAND` accepts anyway). So if you just put the commands in the script directly to cmake list, I will be happy. What I like about it is that there is no redundant shell script file and there are no problems with paths in that script. But this is mostly an improvement, I will merge this PR as long as you stick to what I asked in the first paragraph.

michalvasko (approved):
Great, looks fine now.

michalvasko:
No, I have not merged it yet, I was waiting for the tests to finish!

jktjkt:
sorry, I misinterpreted the notification e-mail, restored now

jktjkt:
...and since it is now closed, I'll delete that branch now. Oh well :).

michalvasko:
Yes, you can do so safely now :)

PR #1128, opened 2022-01-06
Added check to see if the commit timer is not null before deleting the timer

bbergquist0930:
The timer was trying to be deleted even if there was no timer.


michalvasko:
Thanks but I have decided to fix it differently.

PR #1140, opened 2022-02-01
Use sysrepo for NACM

syyyr:
This patch changes Netopeer2 to use the sysrepo version of NACM. There is one problem and that's the test for NACM: if run by the sysrepo superuser, the test is skipped completely, because this user bypasses NACM. When not running as a superuser, some tests are also skipped, because they depend on being run as the superuser.

This needs the sysrepo-nacm PR: https://github.com/sysrepo/sysrepo/pull/2703


src/netconf.c
```diff
@@ -300,7 +301,11 @@ np2srv_rpc_get_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char
 
     /* perform correct NACM filtering */
     sr_session_get_orig_data(session, 1, NULL, (const void **)&username);
-    ncac_check_data_read_filter(&data_get, username);
+    err_info = sr_nacm_check_data_read_filter(&data_get, username);
+    if (err_info) {
+        rc = SR_ERR_INTERNAL;
```
michalvasko, line 306: No need to set your own err code, just use the one in `err_info`. More similar cases follow.

src/main.c
```diff
@@ -448,7 +453,12 @@ np2srv_diff_check_cb(sr_session_ctx_t *session, const struct lyd_node *diff)
     }
 
     sr_session_get_orig_data(session, 1, NULL, (const void **)&user);
-    if ((node = ncac_check_diff(diff, user))) {
+    err_info = sr_nacm_check_diff(diff, user, &node);
```
michalvasko, line 456: We have talked about removing this whole callback, is that still the plan?

syyyr: I think it is, but I'm not 100% sure I'll be able to do that by myself. It seems that the only place it is called from is `sr_changes_notify_store`, which appears to be some sort of a subscription-related function. I would think that the diff check would be called in something like `sr_apply_changes`. Sorry, my understanding of sysrepo code is minimal... :(

michalvasko: That is fine but my first idea, which I have mentioned before, would be to basically just set a flag in the NACM variant of `sr_apply_changes()` and pass it until the point where the callback is called and there just run the code currently in the *netopeer2* diff check callback in case the flag is set, nothing else.

michalvasko: Actually, the user should be a parameter of `sr_nacm_apply_changes()` or whatever the name and also serve as a flag whether to apply NACM or not. All the other NACM functions have this argument so it only makes sense.

PR #1154, opened 2022-02-24
Fix setup error printing

syyyr:
When running tests via ctest, setup failure messages are not shown for
some reason (they are shown when running the test directly). Fix this by
changing the macro to print to stderr.


michalvasko:
No problem, merging.

PR #1155, opened 2022-02-24
Save sysrepo executable env vars after first cmake

syyyr:
Currently, if you're using SYSREPO{CTL,CFG}_EXECUTABLE env vars, you
have to specify it manually when installing Netopeer2 and also when
running tests.

The new code saves the env variable, so that you only have to specify it
when running cmake for the first time.


CMakeLists.txt
```diff
@@ -323,6 +323,22 @@ set(SCRIPT_DIR "${PROJECT_SOURCE_DIR}/scripts/")
...
+endif()
+
+# find sysrepocfg to be used for installation and tests
+if (NOT SYSREPOCFG_EXECUTABLE)
+    find_program(SYSREPOCFG_EXECUTABLE sysrepocfg)
+endif()
+if (NOT SYSREPOCFG_EXECUTABLE)
+    message(FATAL_ERROR "Unable to find sysrepocfg, set SYSREPOCFG_EXECUTABLE manually.")
+endif()
+
```
michalvasko, line 341: Does not seem to me to be the right place for this code. Either put it after `# sysrepo` in the dependency search some lines above or before `if(INSTALL_MODULES)` below, if interpreting it as being needed only for the scripts.

michalvasko (changes requested):
Just a minor problem, otherwise I have nothing against.

CMakeLists.txt
syyyr: It's needed for both tests and scripts, so I put after the find_package stuff

jktjkt (approved):

michalvasko (approved):

michalvasko:
Okay, fine.

Issue #1231, opened 2022-06-17
systemd service not working anymore


Haititi:
Hi,

The `netopeer2-server` binary is in `/usr/sbin` instead of `/usr/bin` since the following commit [24115168e121f75b492f3eb8146794eb7b4ef3f2](https://github.com/CESNET/netopeer2/commit/24115168e121f75b492f3eb8146794eb7b4ef3f2).

It seems that you forgot to update the systemd service file in this commit.
Can you fix the following file [netopeer2-server.service.in](https://github.com/CESNET/netopeer2/blob/devel/service/netopeer2-server.service.in) ?

Moreover, it would be nice to add a `-U` parameter in order to communicate with the server with a socket.
Thanks :)

michalvasko:
Of course, path fixed. About the `-U` parameter, I do not think the service should be listening on any socket. You will have to make some custom changes if you want it to.

PR #1234, opened 2022-07-04
Fix Systemd Unit file netopeer2-server.service

Haititi:
The netopeer2-server binary is in /usr/sbin instead of /usr/bin
since the following commit 2411516.
Moreover, a "-U" parameter is added to be able to communicate
with the service via Unix socket.

This fixes issue #1231.


michalvasko:
Thanks, have not noticed it before and the socket does not really seems appropriate.

Issue #1332, opened 2023-01-05
Segmentation fault (netopeer2-server) in np2srv_rpc_cb() because of err_info fields are corrupted


mblacksee:
Hi Michal,
I've been observing segmantation fault because of fields of `err_info` taken from the user session (user_sess->sess) somehow corrupted in `np2srv_rpc_cb` function. Corresponsing code snippet from the function is as below;

```
/* sysrepo API, use the default timeout or slightly higher than the configured one */
    rc = sr_rpc_send_tree(user_sess->sess, rpc, np2srv.sr_timeout ? np2srv.sr_timeout + 2000 : 0, &output);
    if (rc) {                                                                    
        ERR("Failed to send an RPC (%s).", sr_strerror(rc));                     
                                                                                 
        /* build proper error */                                                 
        sr_session_get_error(user_sess->sess, &err_info);                        
        return np2srv_err_reply_sr(err_info);                                    
    }                                                                            
```

Crash happens in `np2srv_err_reply_sr` function.
In my use case, reason for RPC to fail is that `replay-start-time` is in the feature.

Sample gdb session output is as below;
```
0  0x154e3a14 in strcmp () from /home/murat/mimosa/ws1/dev-C6x-1/mbr-multi/out/projects/c6x/build/c6x/squashfs-root/lib/libc.so.0
#1  0x15359342 in sr_err_get_netconf_error () from /home/murat/mimosa/ws1/dev-C6x-1/mbr-multi/out/projects/c6x/build/c6x/squashfs-root/lib/libsysrepo.so.7
#2  0x000189dc in np2srv_err_nc (err=0x49c28) at /mbr-multi/out/projects/c6x/build/netopeer2-2.1.49/src/main.c:209
#3  0x00018ef6 in np2srv_err_reply_sr (err_info=0x4aad0) at /mbr-multi/out/projects/c6x/build/netopeer2-2.1.49/src/main.c:338
#4  0x000191e2 in np2srv_rpc_cb (rpc=0x4a638, ncs=0x14fb80) at /mbr-multi/out/projects/c6x/build/netopeer2-2.1.49/src/main.c:421
#5  0x153ab448 in nc_ps_poll () from /home/murat/mimosa/ws1/dev-C6x-1/mbr-multi/out/projects/c6x/build/c6x/squashfs-root/lib/libnetconf2.so.3
[...]

PR #1336, opened 2023-01-23
Lock user_sess for err_info access

mblacksee:
Refs #1332

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -146,6 +146,7 @@ np_ps_match_cb(struct nc_session *session, void *cb_data)
     struct np2_user_sess *user_sess;
 
     if (match_data->sr_id) {
+        /* No need to have user session lock for this case */
         user_sess = nc_session_get_data(session);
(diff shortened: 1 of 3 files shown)


michalvasko:
Right, the notifications can always be sent in parallel (only using internal locks in *libnetconf2*) so the `err_info` access can cause data races and even RPCs can be sent in parallel with the notifications. Thanks for discovering the cause of this problem.

src/common.c
```diff
@@ -217,6 +218,10 @@ np_get_user_sess(sr_session_ctx_t *ev_sess, const char *func, struct nc_session
 
     /* user sysrepo session */
     us = nc_session_get_data(ncs);
+    if ((rc = pthread_mutex_lock(&us->lock))) {
```
michalvasko, line 221: To be robust, use a timeout (defined by a macro in `config.h`).

michalvasko (changes requested):

PR #1389, opened 2023-04-24
add commandline option to set PAM Module Name/DIR used for interactive ssh authentication

qupfer:
Hi,
@roytak implemented pam usage for the netconf2 lib. (https://github.com/CESNET/libnetconf2/pull/382).
But it seems there is no way to use it without patching the source of either netconf2 or dependent tools like netopeer2.

So I add the command line option for the netopeer2 server to make it usable without add the name/dir to the sources.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -49,7 +49,7 @@
 #include "log.h"
 #include "netconf_monitoring.h"
 
-struct np2srv np2srv = {.unix_mode = -1, .unix_uid = -1, .unix_gid = -1};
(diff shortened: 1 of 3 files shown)


michalvasko (changes requested):
The code seems fine to me, but please merge the commits into a single one, I have no other problems with this PR.

qupfer:
> The code seems fine to me, but please merge the commits into a single one, I have no other problems with this PR.

done

michalvasko (approved):

PR #1395, opened 2023-05-08
ncc BUGFIX Clean backup dir properly after successful commit

synther:
Hi!

My expectation is that after the confirming commit is issued, backup files (a bunch of jsons and meta file in `/var/netopeer2/confirmed_commit/`) should be cleaned. I observe they are kept and after the server restarts it rollbacks the changes even the confirming commit was successful. 

I made a fix according to my understanding, please check.


michalvasko:
Seems fine, thanks.

PR #1405, opened 2023-05-24
Add support for running netopeer2-server in foreground with syslog

troglobit:
This patch adds support for running `netopeer2-server` under init systems like [Finit](https://github.com/troglobit/finit/), which supervise processes running in the foreground.  Similar to `-d`, but still using syslog.

### DIFF
--- a/doc/netopeer2-server.8
+++ b/doc/netopeer2-server.8
@@ -2,13 +2,13 @@
 .\" groff -man -Tascii netopeer2-server.8
 .\"
 
-.TH "netopeer2-server" 8 "2021-11-10" "Netopeer"
+.TH "netopeer2-server" 8 "2023-05-25" "Netopeer"
 .SH NAME
 netopeer2-server \- NETCONF server daemon build on libnetconf2 with sysrepo datastore
 .
 .SH SYNOPSIS
 .B netopeer2-server
-[\fB-dhV\fP] [\fB-p\fP \fIPATH\fP] [\fB-U\fP[\fIPATH\fP]] [\fB-m\fP \fIMODE\fP] [\fB-u\fP \fIUID\fP]
+[\fB-dFhV\fP] [\fB-p\fP \fIPATH\fP] [\fB-U\fP[\fIPATH\fP]] [\fB-m\fP \fIMODE\fP] [\fB-u\fP \fIUID\fP]
 [\fB-g\fP \fIGID\fP] [\fB-t\fP \fITIMEOUT\fP] [\fB-v\fP \fILEVEL\fP] [\fB-c\fP \fICATEGORY\fP]
 .br
 .
@@ -21,6 +21,9 @@ YANG module data in sysrepo.
 .TP
 .BR "\-d"
 Debug mode (do not daemonize and print verbose messages to stderr instead of syslog).
+.TP
+.BR "\-F"
+Run in foreground (do not daemonize) and log to syslog.
 .TP
 .BR "\-h"
 Display help.
--- a/src/main.c
+++ b/src/main.c
@@ -1078,9 +1078,10 @@ print_version(void)
 static void
 print_usage(char *progname)
 {
-    fprintf(stdout, "Usage: %s [-dhV] [-p PATH] [-U[PATH]] [-m MODE] [-u UID] [-g GID] [-t TIMEOUT] [-x PATH]\n", progname);
+    fprintf(stdout, "Usage: %s [-dFhV] [-p PATH] [-U[PATH]] [-m MODE] [-u UID] [-g GID] [-t TIMEOUT] [-x PATH]\n", progname);
     fprintf(stdout, "          [-v LEVEL] [-c CATEGORY]\n");
     fprintf(stdout, " -d         Debug mode (do not daemonize and print verbose messages to stderr instead of syslog).\n");
+    fprintf(stdout, " -F         Run in foreground, like -d, but log to syslog.\n");
     fprintf(stdout, " -h         Display help.\n");
     fprintf(stdout, " -V         Show program version.\n");
     fprintf(stdout, " -p PATH    Path to pidfile (default path is \"%s\").\n", NP2SRV_PID_FILE_PATH);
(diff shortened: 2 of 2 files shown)


src/main.c
```diff
@@ -1119,7 +1120,7 @@ main(int argc, char *argv[])
 {
     int ret = EXIT_SUCCESS;
     int c, *idx, i;
-    int daemonize = 1, verb = 0;
+    int daemonize = 1, verb = 0, use_syslog = 0;
```
michalvasko, line 1123: If you are adding a flag, use it correctly. By default (with no options), syslog is used so I would set it to `1`. Then, for the `d` option, it should be set to `0`.

src/main.c
```diff
@@ -1152,11 +1153,15 @@ main(int argc, char *argv[])
...
-    while ((c = getopt(argc, argv, "dhVp:f:U::m:u:g:n:i:t:x:v:c:")) != -1) {
+    while ((c = getopt(argc, argv, "dhVp:f:FU::m:u:g:n:i:t:x:v:c:")) != -1) {
         switch (c) {
         case 'd':
             daemonize = 0;
             break;
+	case 'F':
+            daemonize = 0;
+            use_syslog = 1;
+            break;
```
michalvasko, line 1164: Note that the cases are ordered the same way as the options in `getopt`, please stick to that order.

src/main.c
```diff
@@ -1319,6 +1324,10 @@ main(int argc, char *argv[])
         np2_stderr_log = 0;
     }
 
+    /* use syslog even if running in foreground */
+    if (use_syslog == 1)
+        np2_stderr_log = 0;
```
michalvasko, line 1329: We do not allow `if` without braces in our coding style.

michalvasko (changes requested):
Okay, this can be merged after some changes. Also, change the base branch to `devel`, we do not accept PRs to `master`.

troglobit:
Will do, thanks for the swift review!

src/main.c
troglobit: Fixed in updated PR (force-push)

src/main.c
troglobit: Fixed in updated PR (force-push)

troglobit:
> Okay, this can be merged after some changes. Also, change the base branch to `devel`, we do not accept PRs to `master`.

Fixed.

src/main.c
```diff
@@ -1199,6 +1201,10 @@ main(int argc, char *argv[])
         case 'f':
             np2srv.server_dir = optarg;
             break;
+	case 'F':
+            daemonize = 0;
+            use_syslog = 1;
```
michalvasko, line 1206: Redundant now.

src/main.c
```diff
@@ -1319,6 +1325,11 @@ main(int argc, char *argv[])
         np2_stderr_log = 0;
     }
 
+    /* use syslog even if running in foreground */
```
michalvasko, line 1328: Inaccurate now, it is not only about the `-F` parameter.

michalvasko (changes requested):
Almost there but I have realized that you should probably also change the man page `doc/netopeer2-server.8`.

src/main.c
troglobit: OK

src/main.c
```diff
@@ -1319,6 +1325,11 @@ main(int argc, char *argv[])
         np2_stderr_log = 0;
```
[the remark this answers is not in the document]
troglobit: Argh, obviously. Sorry about that, I don't even need the `use_syslog` flag. Fixing!

src/main.c
troglobit: Fixing ...

troglobit:
> Almost there but I have realized that you should probably also change the man page `doc/netopeer2-server.8`.

Of course, will fix that too.  New update coming later tonight (CET).

troglobit:
There, finally.  This last changeset actually doesn't look half bad.  Note, I decided to move the case and group it closer to `-d`, which the new option has a lot in common with.  Because I noticed all other non-argument options were sorted first.

doc/netopeer2-server.8
```diff
@@ -8,7 +8,7 @@ netopeer2-server \- NETCONF server daemon build on libnetconf2 with sysrepo data
 .
 .SH SYNOPSIS
 .B netopeer2-server
-[\fB-dhV\fP] [\fB-p\fP \fIPATH\fP] [\fB-U\fP[\fIPATH\fP]] [\fB-m\fP \fIMODE\fP] [\fB-u\fP \fIUID\fP]
+[\fB-dFhV\fP] [\fB-p\fP \fIPATH\fP] [\fB-U\fP[\fIPATH\fP]] [\fB-m\fP \fIMODE\fP] [\fB-u\fP \fIUID\fP]
```
michalvasko, line 11: Please update the last change on line 5 as well for it to be obvious that is is up-to-date.

michalvasko (changes requested):
Right, code changes are really simple now so just this one last detail and it should be good to go.

doc/netopeer2-server.8
troglobit: Heh, of course! :-)

troglobit:
> Right, code changes are really simple now so just this one last detail and it should be good to go.

Aaaaand done :smiley::+1:

michalvasko (approved):

src/main.c
```diff
@@ -1152,11 +1153,15 @@ main(int argc, char *argv[])
     np2srv.server_dir = SERVER_DIR;
 
     /* process command line options */
-    while ((c = getopt(argc, argv, "dhVp:f:U::m:u:g:n:i:t:x:v:c:")) != -1) {
+    while ((c = getopt(argc, argv, "dFhVp:f:U::m:u:g:n:i:t:x:v:c:")) != -1) {
         switch (c) {
         case 'd':
             daemonize = 0;
             break;
+	case 'F':
```
michalvasko, line 1161: Is this a TAB? If so, it is the reason why the tests failed.

troglobit: Indeed, it was a TAB, really thought I had cleaned that up. Sorry, fixing!

PR #1418, opened 2023-06-13
pubkey authentication for non system user

qupfer:
Hi,

if `NP2SRV_SSH_AUTHORIZED_KEYS_ARG_IS_USERNAME` is set, then there is no need that the netopeer client is a system user. But the current version fails, if the user does not exist in the user database. 

This PR skips that, if NP2SRV_SSH_AUTHORIZED_KEYS_ARG_IS_USERNAME is set. 

Kind regards
Henning

### DIFF
--- a/src/netconf_server_ssh.c
+++ b/src/netconf_server_ssh.c
@@ -82,6 +82,32 @@ cleanup:
     return rc;
 }
 
+static int
+np2srv_validate_posix_username(const char *username)
+{
+    /* use POSIX username definition
+     * https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_437 */
+
+    /* not empty */
+    if (strlen(username) == 0) {
+        return -1;
+    }
+
+    /* no hyphen as first char */
+    if (username[0] == '-') {
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


src/netconf_server_ssh.c
```diff
@@ -92,19 +92,48 @@ np2srv_pubkey_auth_cb(const struct nc_session *session, ssh_key key, void *UNUSE
...
 
     errno = 0;
     pwd = getpwnam(username);
-    if (!pwd) {
+    if (!NP2SRV_SSH_AUTHORIZED_KEYS_ARG_IS_USERNAME && !pwd) {
         ERR("Failed to find user entry for \"%s\" (%s).", username, errno ? strerror(errno) : "User not found");
         goto cleanup;
     }
 
+    /* if the user is not a system user, we need at least a bit of validation */
```
michalvasko, line 106: Please put the validation into a separate function.

michalvasko (changes requested):
Fine, I have nothing against the change but please run `uncrustify` on the code, there are coding style errors.

src/netconf_server_ssh.c
qupfer: done.

michalvasko (approved):

PR #1421, opened 2023-06-15
compat: fix `pthread_mutex_timedlock`

jktjkt:
## refactor: compat: simplify waiting
    
This code looked a bit fishy:
    
- There's no need to check up front whether we're dividing a zero or non-zero number by some constant.
    
- The conversion to milliseconds is actually not needed at all because it gets immediately converted to back to nanoseconds.


jktjkt:
I accidentally left this one marked as a draft; you can merge it now (and please also https://github.com/CESNET/libnetconf2/pull/424).

PR #1431, opened 2023-07-03
Configured subscription integration

jeremie6wind:
Hello,
This time, add udp notification code from https://github.com/network-analytics/udp-notif-c-collector.
Build the code with the rest of the project.
Best regards
Jeremie

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ ExternalProject_Add(udp-notif-c-collector
     GIT_REPOSITORY https://github.com/network-analytics/udp-notif-c-collector
     GIT_TAG main
     BUILD_IN_SOURCE 1
+    PATCH_COMMAND patch -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/patches/0001-feat-add-local-address-to-sender.patch
     CONFIGURE_COMMAND ./bootstrap && ./configure --prefix=${UDP_NOTIF_BIN}
     BUILD_COMMAND make
     INSTALL_COMMAND make install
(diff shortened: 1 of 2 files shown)


michalvasko (commented):
So you added the library as an external project but also included it in the repository? You can download it first and then it does not have to be directly included. Or skip the external project altogether and just add the source files to *netopeer2* sources to avoid redundant static library and CMake code.

jeremie6wind:
Solution 1:
Using the external project with the clone means that if we want to modify udp-notif-c-collector, then we need to push to the original repository to modify it. IMHO, it is not very practical.

Solution2:
Adding the source files leads to building issues, that requires adding _GNU_SOURCE
to the udp-notif-c-collector source files. Also unyte_version.h is normally generated at build time. Finally the indentation is not the same as Netopeer2.

michalvasko:
> Solution 1:
Using the external project with the clone means that if we want to modify udp-notif-c-collector, then we need to push to the original repository to modify it. IMHO, it is not very practical.

I would expect that if there are improvements to be made you would want to push them upstream but you can also apply patches to the downloaded source.

> Solution2:
Adding the source files leads to building issues, that requires adding _GNU_SOURCE
to the udp-notif-c-collector source files. Also unyte_version.h is normally generated at build time. Finally the indentation is not the same as Netopeer2.

I see. Then I suppose this is not an option, too bad the project is also using `autotools` instead of `cmake`.

jeremie6wind:
the 2 commits  just pushed add the external project in the CMakeLists.txt and a patch to set the source address of udp notifications.

michalvasko (approved):
Thanks for that, looks much cleaner.

PR #1469, opened 2023-09-14
netopeer2 with new libnetconf2 API

Roytak:
Initial changes for netopeer2 in order to use new libnetconf2 API.

Some of the most notable changes include:
- removal of configuration callbacks, it's handled by ln2
- removal of some redundant command line opts
- removal of some np2's YANG modules, they are loaded via ln2's schema dir
- the SSH and TLS options have been merged into one, that is SSH_TLS


tests/np_test.c
```diff
@@ -221,9 +232,7 @@ np_glob_setup_np2(void **state, const char *test_name, const char **modules)
         close(fd);
 
         /* exec server listening on a unix socket */
```
michalvasko, line 234: There is a UNIX socket but not set up here so adjust the comment.

michalvasko (changes requested):

scripts/common.sh
```diff
@@ -0,0 +1,94 @@
...
+
+    if [ -z "$SYSREPOCFG" ]; then
+    	echo "$0: Unable to find sysrepocfg executable." >&2
+        exit 1
+    fi
+}
+
+# get path to the sysrepoctl executable
+function SYSREPOCTL_GET_PATH() {
+	if [ -n "$SYSREPOCTL_EXECUTABLE" ]; then
```
michalvasko, line 60: Indent.

scripts/merge_config.sh
```diff
@@ -2,20 +2,69 @@
...
+            IDX=$((IDX + 1))
+        fi
+    done < "$AUTHORIZED_KEYS_FILE"
+
+    # append the ending tags
+    AUTH_CONFIG="${AUTH_CONFIG}
+                            </inline-definition>
+                        </public-keys>"
+
+    echo "Creating user \"${CURRENT_USER}\" that can authenticate with a key pair from his authorized_keys..."
```
michalvasko, line 60: The message (and the one below) is a bit inaccurate, the user is not being created, only added into the server configuration.

scripts/remove.sh
```diff
@@ -1,54 +1,31 @@
 #!/usr/bin/env bash
 # {% raw %} - jinja2 command to not process "{#" in the script
```
michalvasko, line 2: This can be removed, obsolete now. At the script end, too.

src/netconf_server.c
```diff
@@ -32,517 +35,153 @@
...
 
-    return SR_ERR_OK;
+    return 0;
 }
 
-/* /ietf-netconf-server:netconf-server/listen/endpoint/ * /tcp-server-parameters */
 int
-np2srv_endpt_tcp_params_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UNUSED(module_name),
-        const char *xpath, sr_event_t UNUSED(event), uint32_t UNUSED(request_id), void *UNUSED(private_data))
+np2srv_pubkey_auth_cb(const struct nc_session *session, ssh_key key, void *UNUSED(user_data))
```
michalvasko, line 84: Server (libnetconf2) should use the authorized keys configured for the client, not read them from the system or am I missing something?

tests/np_test.c
```diff
@@ -167,6 +175,10 @@ np_glob_setup_np2(void **state, const char *test_name, const char **modules)
         SETUP_FAIL_LOG;
         return 1;
     }
+    if (unsetenv("LN2_MODULE_DIR")) {
```
michalvasko, line 178: Missing unset for `NP2_SCRIPTS_DIR`.

michalvasko (changes requested):

Roytak:
Figured there is no need for the files `netconf_server.c` and `netconf_server.h` if there's only one function definition (the libnetconf2 callback) in them, so removed them.

michalvasko (approved):

PR #1550, opened 2024-03-12
subscriptions UPDATE merge operational data

jeremie6wind:
An external tool that implements configured subscriptions can fill other leaves by subscribing the operational data with the SR_SUBSCR_OPER_MERGE flag. For instance, the
configured-subscription-state field defined in the RFC 8639 Link:https://datatracker.ietf.org/doc/html/rfc8639.


src/main.c
```diff
@@ -943,7 +943,12 @@ server_data_subscribe(void)
 
     /* operational data */
     SR_OPER_SUBSCR(mod_name, "/ietf-subscribed-notifications:streams", srsn_oper_data_streams_cb);
-    SR_OPER_SUBSCR(mod_name, "/ietf-subscribed-notifications:subscriptions", np2srv_oper_sub_ntf_subscriptions_cb);
+    rc = sr_oper_get_subscribe(np2srv.sr_sess, mod_name, "/ietf-subscribed-notifications:subscriptions",
+            srsn_oper_data_streams_cb, NULL, SR_SUBSCR_OPER_MERGE, &np2srv.sr_data_sub);
```
michalvasko, line 947: You have most likely unintentionally changed the callback.

michalvasko (changes requested):

src/main.c
jeremie6wind: Yes my mistake.

michalvasko:
Okay, no problem, makes sense.

PR #1552, opened 2024-03-15
Implementing netconf-confirmed-commit

lePici:
Notification for YANG module ietf-netconf-notifications.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -525,6 +525,92 @@ cleanup:
     return rc;
 }
 
+int
+np_send_notif_confirmed_commit(const struct nc_session *session, sr_session_ctx_t *sr_session, enum np_cc_event event,
+        uint32_t cc_timeout, uint32_t sr_timeout)
+{
+    int rc = 0, r;
+    const struct ly_ctx *ly_ctx;
+    const struct lys_module *mod;
+    struct lyd_node *notif = NULL;
+    char num32[11]; /* max bytes for 32-bit unsigned number + \0 */
+    char *value;
+
+    /* get context */
+    if (session) {
+        ly_ctx = nc_session_get_ctx(session);
+    } else {
+        assert(event == NP_CC_TIMEOUT);
+        ly_ctx = sr_session_acquire_context(sr_session);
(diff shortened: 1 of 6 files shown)


src/common.c
```diff
@@ -371,6 +371,234 @@ np_ntf_send(struct nc_session *ncs, const struct timespec *timestamp, struct lyd
...
+{
+    int rc = 0, r;
+    const struct ly_ctx *ly_ctx;
+    const struct lys_module *mod;
+    struct lyd_node *notif = NULL;
+    char num32[11]; /* max bytes for 32-bit unsigned number + \0 */
+    char *value;
+
+    ly_ctx = nc_session_get_ctx(session);
+    if ((mod = ly_ctx_get_module_implemented(ly_ctx, "ietf-netconf-notifications"))) {
```
michalvasko, line 461: This `if` could be negated and lead to `cleanup`, would avoid having the whole function in the condition (which I prefer to not do). Also, if you have the `mod`, you should use it and call `lyd_new_inner()` instead of `lyd_new_path()`, it is unused currently.

src/netconf_confirmed_commit.c
```diff
@@ -884,6 +891,10 @@ np2srv_confirmed_commit_cb(sr_session_ctx_t *session, const struct lyd_node *inp
         /* there is already a pending confirmed commit, keep its backup, but the timeout will be reset */
         timer_delete(commit_ctx.timer);
         commit_ctx.timer = 0;
+
+        /* send notification about extending timeout for confirmed-commits */
+        np_send_notif_confirmed_commit(nc_sess, session, NP_CC_EXTEND, timeout, 0);
```
michalvasko, line 896: Just to make the code a bit more readable I would just set the flag `timeout_changed` here and then later, based on that, send one or the other notification.

michalvasko (changes requested):
Looks almost perfect, just one minor comment and also please reference issue 1548 in the commit adding the new notification.

michalvasko (approved):

PR #1606, opened 2024-07-10
Generate certificate expiration notifications

Roytak:
These notifications will only be generated, if the `ietf-crypto-types` YANG feature `certificate-expiration-notification` is enabled.


Roytak:
Should this new capability be mentioned anywhere in `netopeer2`? Did not find an adequate place for it and not sure if mentioning it only in `libnetconf2` is sufficient.

src/main.c
```diff
@@ -505,6 +505,86 @@ np2srv_sm_oper_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char
...
 }
 
+#ifdef NC_ENABLED_SSH_TLS
+
+void
+np2srv_cert_exp_notif_cb(const char *expiration_time, const char *xpath, void *user_data)
+{
+    sr_session_ctx_t *sr_sess = user_data;
+    const struct ly_ctx *ly_ctx = NULL;
+    int r;
```
michalvasko, line 515: Please stick to the convention, use an initialized `rc` (`ret` is also fine) for function return value, `r` is for temporary returns.

michalvasko (changes requested):
I am guessing you left out some new files or something.

src/main.c
Roytak: Added a new flag and renamed `r` to `rc`.

src/main.c
```diff
@@ -505,6 +505,88 @@ np2srv_sm_oper_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char
...
+    if (ly_ctx) {
+        sr_release_context(np2srv.sr_conn);
+    }
+    if (stop_thread) {
+        nc_server_notif_cert_expiration_thread_stop(1);
+    }
+}
+
+static int
+np2srv_start_cert_exp_notif_thread()
```
michalvasko, line 550: In function declarations we always specify at least `void` if there are no parameters, use empty parentheses only when calling such a function.

src/main.c
```diff
@@ -505,6 +505,88 @@ np2srv_sm_oper_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char
     return rc;
 }
 
+#ifdef NC_ENABLED_SSH_TLS
+
+void
+np2srv_cert_exp_notif_cb(const char *expiration_time, const char *xpath, void *user_data)
```
michalvasko, line 511: Should be `static` and missing doxygen then (for callbacks with typedefs only a short brief description should be supplied).

src/main.c
```diff
@@ -505,6 +505,88 @@ np2srv_sm_oper_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char
...
+    if (ly_ctx) {
+        sr_release_context(np2srv.sr_conn);
+    }
+    if (stop_thread) {
+        nc_server_notif_cert_expiration_thread_stop(1);
+    }
+}
+
+static int
+np2srv_start_cert_exp_notif_thread(void)
```
michalvasko, line 550: Missing doxygen.

michalvasko (changes requested):

src/main.c
Roytak: Added both.

michalvasko (approved):

PR #1620, opened 2024-08-08
Move SERVER_DIR to /var/lib in rpm

### LINKED ISSUE #1619: Failed changing permissions of directory "/usr/lib64/netopeer2-server" (Read-only file system)
Hey,

I encountered a problem with netopeer2-server-2.1.71-1.el8.x86_64

I don't know if this is an issue with this repo or someone elses build configuration, as I have a hard time finding out who built this package and what the configuration was (see SERVER_DIR below).
But maybe you can enlighten me.

When using a request that changes the candidate target and then commits:

```xml
<rpc xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="101">
  <edit-config>
    <target>
      <candidate/>
    </target>
    <config>
      <system xmlns="urn:ietf:params:xml:ns:yang:ietf-system">
        <hostname>foobar</hostname>
      </system>
    </config>
  </edit-config>
</rpc>

<rpc  xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="102">
  <commit>
    <confirmed/>
  </commit>
</rpc>
```

The following error is logged:

```
Failed changing permissions of directory "/usr/lib64/netopeer2-server" (Read-only file system)
```

I managed to track down the error to the following line in the systemd unit file:

https://github.com/CESNET/netopeer2/blob/1e639fe18a70b06111ad9c3ea261f27794f25f0e/distro/pkg/rpm/netopeer2-server.service#L11

which mounts /usr read only:

```
$ man systemd.exec
...
       ProtectSystem=
           Takes a boolean argument or the special values "full" or "strict". If true, mounts the /usr and /boot directories read-only 
           processes invoked by this unit.
```

in combination with SERVER_DIR being `/usr/lib64/netopeer2-server` in that build (I don't know if that comes from your repo or the distro build config), the directory cannot be written/created/chmodded.

This:
https://github.com/CESNET/netopeer2/blob/1e639fe18a70b06111ad9c3ea261f27794f25f0e/CMakeLists.txt#L117

looks like it should be set to /var/netopeer2 instead, but it seems like it is not..

I guess 

https://github.com/CESNET/netopeer2/blob/1e639fe18a70b06111ad9c3ea261f27794f25f0e/distro/pkg/rpm/netopeer2.spec#L68

is the culprit, maybe this should not use `_libdir`, but some `/var`-like directory, like in the debian build:

https://github.com/CESNET/netopeer2/blob/1e639fe18a70b06111ad9c3ea261f27794f25f0e/distro/pkg/deb/rules#L21

Changing server_dir via the -f option (or, less safe, changing ProtectSystem from yes to no) is a usable workaround for now.

marvinthepa:
Fixes #1619.

### DIFF
--- a/distro/pkg/rpm/netopeer2.spec
+++ b/distro/pkg/rpm/netopeer2.spec
@@ -65,13 +65,13 @@ a single established NETCONF session.
 %cmake -DCMAKE_BUILD_TYPE=RELWITHDEBINFO \
        -DSYSREPO_SETUP=OFF \
        -DPIDFILE_PREFIX=/run \
-       -DSERVER_DIR=%{_libdir}/netopeer2-server
+       -DSERVER_DIR=%{_sharedstatedir}/netopeer2-server
 %cmake_build
 
 %install
 %cmake_install
 install -D -p -m 0644 %{SOURCE2} %{buildroot}%{_unitdir}/netopeer2-server.service
-mkdir -p -m=700 %{buildroot}%{_libdir}/netopeer2-server
+mkdir -p -m=700 %{buildroot}%{_sharedstatedir}/netopeer2-server
 
 %post server
 set -e
@@ -102,7 +102,7 @@ set -e
 %{_sysconfdir}/pam.d/netopeer2.conf
 %dir %{_datadir}/yang/modules/netopeer2/
 %dir %{_datadir}/netopeer2/
-%dir %{_libdir}/netopeer2-server/
+%dir %{_sharedstatedir}/netopeer2-server/
 
 %files cli
 %license LICENSE


distro/pkg/rpm/netopeer2.spec
```diff
@@ -65,13 +65,12 @@ a single established NETCONF session.
 %cmake -DCMAKE_BUILD_TYPE=RELWITHDEBINFO \
        -DSYSREPO_SETUP=OFF \
        -DPIDFILE_PREFIX=/run \
-       -DSERVER_DIR=%{_libdir}/netopeer2-server
+       -DSERVER_DIR=%{_var}/lib/netopeer2-server
```
jktjkt, line 68: My RPM manual says that it's `%{_sharedstatedir}` for (an equivalent of) `/var/lib`.

distro/pkg/rpm/netopeer2.spec
```diff
@@ -65,13 +65,12 @@ a single established NETCONF session.
...
 %cmake_build
 
 %install
 %cmake_install
 install -D -p -m 0644 %{SOURCE2} %{buildroot}%{_unitdir}/netopeer2-server.service
-mkdir -p -m=700 %{buildroot}%{_libdir}/netopeer2-server
```
jktjkt, line 74: Don't you need to `mkdir %{_sharedstatedir}/netopeer2-server` somewhere, and add it to the `%files` section as well? Or will the server create it at runtime?

jktjkt (changes requested):

distro/pkg/rpm/netopeer2.spec
marvinthepa: The server did create the directory at runtime in my experience, which matches the code:

https://github.com/CESNET/netopeer2/blob/1e639fe18a70b06111ad9c3ea261f27794f25f0e/src/netconf_confirmed_commit.c#L207-L217

As `/var` exists for services to write to at runtime, I thought this behavior was okay.

But I don't mind creating it and adding it to files either. Let me know what you prefer.

marvinthepa: I checked the "docker.io" package on Ubuntu, and it does not include `/var/lib/docker`, but the daemon creates it at runtime.
I don't know docker can be considered as a standard to adhere to, though..

jktjkt: Since this is a `.spec` file, that means RPM, so I suggest to look at RPM-based distros. Fedora is the most common one (at least regarding various distro-specific rules).

marvinthepa: You are absolutely right, RPM based Distros seem to follow a different convention, I found several packages that do include their /var/lib directories in the RPM. Commit updated.

michalvasko (approved):

michalvasko:
Thanks, no problem. I have just changed the target branch to `devel`, we do not accept PRs to `master`. Will get there on the next release.

Issue #1556, opened 2024-04-03
Issues in specific Error tags generation 


Omnamasivaya26:
As per RFC6241, I am supposed to push these errors using edit-config in Netopeer2, but I'm unable to get the below error tags. Instead, I'm getting an 'operation-failed' error-tag. I have checked with other Netopeer versions, including version 2.1.36, where it described the fix for Netconf error tags, but still, I'm unable to see the expected error-tag. Any guidance would be much appreciated.

1.error-tag: missing-attribute
error-type: rpc, protocol, application
error-severity: error error-info: <bad-attribute> : name of the missing attribute.
<bad-element> : name of the element that contains the unexpected attribute
Description: An expected attribute is missing.

2.error-tag: unknown-attribute
error-type: rpc, protocol, application
error-severity: error error-info: <bad-attribute> : name of the unexpected attribute
<bad-element> : name of the element that contains the unexpected attribute
Description: An unexpected attribute is present.

3.error-tag: missing-element
error-type: protocol, application
error-severity: error
error-info: <bad-element> : name of the missing element
Description: An expected element is missing.

4.error-tag: resource-denied
error-type: transport, rpc, protocol, application
error-severity: error
error-info: none
Description: Request could not be completed because of insufficient resources.

5.error-tag: malformed-message
error-type: rpc
error-severity: error
error-info: none
Description: A message could not be handled because it failed to
be parsed correctly. For example, the message is not
well-formed XML or it uses an invalid character set.

michalvasko:
Describe the exact use-cases for all of these errors and I can add them if it makes sense.

Omnamasivaya26:
> Describe the exact use-cases for all of these errors and I can add them if it makes sense.

To validate or modify the Netconf server by pushing XMLs is the traditional approach, and the <error-tag> is essential for pinpointing exact issues within the XML.For example, If an XML includes an unknown or missing attribute, it should accurately trigger the specified <error-tag>. This precision is vital for ensuring proper validation of the Netconf server. Therefore, the error tags I have mentioned appear to be crucial for validation

1)While using edit-config, If the xml  have an expected attribute missing then it should throw the missing-attribute error tag

2)Similarly, If the xml  have an undefined attribute then it should throw the unknown-attribute error tag

3)Similarly, If the xml  have an expected element missing then it should throw the missing-element error tag

4)Similarly, If the xml request failed due to insufficient resource then it should throw the resource-denied error tag

[...]

PR #1636, opened 2024-09-04
NETCONF error tags

lePici:
The last partial solution for CESNET/netopeer2#1556.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -250,6 +250,64 @@ error:
 }
 
 /**
+ * @brief Find the nth substring delimited by quotes.
+ *
+ * For example: abcd"ef"ghij"kl"mn -> index 0 is "ef", index 1 is "kl".
+ *
+ * @param[in] msg Input string with quoted substring.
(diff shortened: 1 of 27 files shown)


src/main.c
```diff
@@ -249,6 +249,63 @@ np2srv_err_nc(sr_error_info_err_t *err)
...
+ * @brief Find the nth substring delimited by quotes.
+ *
+ * For example: abcd"ef"ghij"kl"mn -> rank 1 is "ef", rank 2 is "kl".
+ *
+ * @param[in] msg Input string with quoted substring.
+ * @param[in] rank Number starting from 1 specifying the nth substring.
+ * @return Copied nth substring without quotes.
+ */
+static char *
+np2srv_err_reply_get_quoted_string(const char *msg, uint32_t rank)
```
michalvasko, line 262: Again, `rank` is not really a good word, `index` seems more accurate.

michalvasko (changes requested):

michalvasko (approved):

PR #1671, opened 2024-11-22
Build and new TLS keys

laaurii00:


michalvasko:
I do not understand this PR. The `build` directory is meant to be created and generated by every user so it should not be a part of the repository. And I do not know what you are doing with the TLS keys.

laaurii00:
Sorry, I was trying to do a fork, you can dismiss this PR. Regarding TLS keys, I was trying to generate new keys and understand how to load them into the server-client.

PR #1684, opened 2024-12-19
cli: make the CLI prompt configurable

hthiery:


michalvasko (changes requested):
Okay, if you find it useful for some reason, but please reorganize the files a bit. No reason for there to be more configurable files, so you can just merge `cli_version.h.in` and `cli_prompt.h.in` to `cli_config.h.in`. Also, please call the macro `NC_CLI_PROMPT` since it is no longer local and it was a bad name anyway.

hthiery:
> Okay, if you find it useful for some reason, but please reorganize the files a bit. No reason for there to be more configurable files, so you can just merge `cli_version.h.in` and `cli_prompt.h.in` to `cli_config.h.in`. Also, please call the macro `NC_CLI_PROMPT` since it is no longer local and it was a bad name anyway.

What do you mean with "call the macro `NC_CLI_PROMPT`"?
- Do you mean I should change the cmake config option to NC_CLI_PROMPT?
- or change the name PROMPT to NC_CLI_PROMPT?

michalvasko:
I meant the macro `PROMPT` to be called `NC_CLI_PROMPT`. CMake var `CLI_PROMPT` can stay as it is.

michalvasko (approved):

hthiery:
indeed .. I used master :-/

michalvasko (approved):

PR #1696, opened 2025-01-31
server UPDATE make thread and poll configurable

jeremie6wind:
Since the commit in the link, some performance issues could be observed. Make the thread count and sleep times configurable to avoid this.

Link: https://github.com/CESNET/netopeer2/commit/5391a73bf27b4878f981bf8a474e206cd2d212c3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,8 @@ option(NETOPEER2_LIB_TESTS "Include server tests in netopeer2 library" ON)
 set(SSH_AUTHORIZED_KEYS_FORMAT "%h/.ssh/authorized_keys" CACHE STRING "sshd-like pattern (with '%h', '%u', '%U') for determining path to users' SSH authorized_keys file.")
 set(THREAD_COUNT 3 CACHE STRING "Number of threads accepting new sessions and handling requests")
 set(POLL_IO_TIMEOUT 10 CACHE STRING "Timeout in milliseconds of polling sessions for new data. It is also used for synchronization of low level IO such as sending a reply while a notification is being sent")
+set(PS_BACKOFF_SLEEP 30 CACHE STRING "Timeout in milliseconds of polling sessions backoff sleep")
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -78,8 +78,16 @@ option(BUILD_NETOPEER2_LIB "Build netopeer2 library with sysrepo YANG module set
...
 option(NETOPEER2_LIB_TESTS "Include server tests in netopeer2 library" ON)
 set(SSH_AUTHORIZED_KEYS_FORMAT "%h/.ssh/authorized_keys" CACHE STRING "sshd-like pattern (with '%h', '%u', '%U') for determining path to users' SSH authorized_keys file.")
-set(THREAD_COUNT 3 CACHE STRING "Number of threads accepting new sessions and handling requests")
-set(POLL_IO_TIMEOUT 10 CACHE STRING "Timeout in milliseconds of polling sessions for new data. It is also used for synchronization of low level IO such as sending a reply while a notification is being sent")
+if(NOT THREAD_COUNT)
+    set(THREAD_COUNT 3 CACHE STRING "Number of threads accepting new sessions and handling requests")
+endif()
+if(NOT POLL_IO_TIMEOUT)
+    set(POLL_IO_TIMEOUT 10 CACHE STRING "Timeout in milliseconds of polling sessions for new data. It is also used for synchronization of low level IO such as sending a reply while a notification is being sent")
+endif()
```
michalvasko, line 86: I do not see the reason for these changes, cached strings can be changed during compilation and `set` is used only as the default.

CMakeLists.txt
```diff
@@ -78,8 +78,16 @@ option(BUILD_NETOPEER2_LIB "Build netopeer2 library with sysrepo YANG module set
...
+if(NOT PS_BACKOFF_SLEEP)
+    set(PS_BACKOFF_SLEEP 30 CACHE STRING "Timeout in milliseconds of polling sessions backoff sleep")
+endif()
```
michalvasko, line 89: No problem with the new option but, again, I think the `if` is redundant.

michalvasko (changes requested):

michalvasko (approved):

PR #1715, opened 2025-03-31
implement ietf-system-capabilities and ietf-notification-capabilities

jeremie6wind:
This patch is an implementation of RFC9696
Link: https://datatracker.ietf.org/doc/html/rfc9196

netopeer2 can return the operational data of /ietf-system-capabilities:system-capabilities
which contains the limit values of several parameters.

max-nodes-per-update: there is no limit on the number of node to send inside a notif in sysrepo. The max unsigned integer value is set.
periodic-notifications-supported and on-change-supported: The publisher is capable of sending notifications for 'config false' and 'config true' nodes.
minimum-update-period and minimum-dampening-period: The minimum value is 0.
supported-excluded-change-type: All types of changes can be excluded.

A sysrepocfg command would give this result:

`# sysrepocfg -f json -x "/ietf-system-capabilities:system-capabilities" -v2 -X -d operational`

```
{
  "ietf-system-capabilities:system-capabilities": {
    "datastore-capabilities": [
      {
        "datastore": "ietf-datastores:startup",
        "per-node-capabilities": [
          {
            "ietf-notification-capabilities:subscription-capabilities": {
              "max-nodes-per-update": 4294967295,
              "periodic-notifications-supported": "config-changes state-changes",
              "minimum-update-period": 0,
              "on-change-supported": "config-changes state-changes",
              "minimum-dampening-period": 0,
              "supported-excluded-change-type": [
                "all"
              ]
            }
          }
        ]
      },
      {
        "datastore": "ietf-datastores:running",
        "per-node-capabilities": [
          {
            "ietf-notification-capabilities:subscription-capabilities": {
              "max-nodes-per-update": 4294967295,
              "periodic-notifications-supported": "config-changes state-changes",
              "minimum-update-period": 0,
              "on-change-supported": "config-changes state-changes",
              "minimum-dampening-period": 0,
              "supported-excluded-change-type": [
                "all"
              ]
            }
          }
        ]
      },
      {
        "datastore": "ietf-datastores:candidate",
        "per-node-capabilities": [
          {
            "ietf-notification-capabilities:subscription-capabilities": {
              "max-nodes-per-update": 4294967295,
              "periodic-notifications-supported": "config-changes state-changes",
              "minimum-update-period": 0,
              "on-change-supported": "config-changes state-changes",
              "minimum-dampening-period": 0,
              "supported-excluded-change-type": [
                "all"
              ]
            }
          }
        ]
      },
      {
        "datastore": "ietf-datastores:operational",
        "per-node-capabilities": [
          {
            "ietf-notification-capabilities:subscription-capabilities": {
              "max-nodes-per-update": 4294967295,
              "periodic-notifications-supported": "config-changes state-changes",
              "minimum-update-period": 0,
              "on-change-supported": "config-changes state-changes",
              "minimum-dampening-period": 0,
              "supported-excluded-change-type": [
                "all"
              ]
            }
          }
        ]
      }
    ],
    "ietf-notification-capabilities:subscription-capabilities": {
      "max-nodes-per-update": 4294967295,
      "periodic-notifications-supported": "config-changes state-changes",
      "minimum-update-period": 0,
      "on-change-supported": "config-changes state-changes",
      "minimum-dampening-period": 0,
      "supported-excluded-change-type": [
        "all"
      ]
    }
  }
}

```

### DIFF
--- a/modules/ietf-notification-capabilities@2022-02-17.yang
+++ b/modules/ietf-notification-capabilities@2022-02-17.yang
@@ -1,0 +1,262 @@
+module ietf-notification-capabilities {
+  yang-version 1.1;
+  namespace
+    "urn:ietf:params:xml:ns:yang:ietf-notification-capabilities";
+  prefix notc;
+
+  import ietf-yang-push {
+    prefix yp;
+    description
+      "This module requires ietf-yang-push to be implemented.";
+    reference
+      "RFC 8641: Subscription to YANG Notifications for
+       Datastore Updates";
+  }
+  import ietf-system-capabilities {
+    prefix sysc;
+    description
+      "This module requires ietf-system-capabilities to be
+       implemented.";
+    reference
+      "RFC 9196: YANG Modules Describing Capabilities for Systems
+       and Datastore Update Notifications";
+  }
+
+  organization
+    "IETF NETCONF (Network Configuration) Working Group";
+  contact
+    "WG Web:   <https://datatracker.ietf.org/wg/netconf/>
+     WG List:  <mailto:netconf@ietf.org>
+
+     Editor:   Balazs Lengyel
+               <mailto:balazs.lengyel@ericsson.com>";
+  description
+    "This module specifies publisher capabilities related to
+     YANG-Push (RFC 8641).
+
+     The module contains:
+
+     - a specification of the data nodes that support 'on-change' or
+     'periodic' notifications.
+
+     - capabilities related to the throughput of notification data
+     that the publisher can support.  (Note that for a specific
+     subscription, the publisher MAY allow only longer periods
+     or smaller updates depending on, e.g., actual load conditions.)
+
+     Capability values can be specified at the system/publisher
+     level, at the datastore level, or for specific data nodes of
+     a specific datastore (and their contained subtrees), as defined
+     in the ietf-system-capabilities module.
+
+     If different data nodes covered by a single subscription
+     have different values for a specific capability, then using
+     values that are only acceptable for some of these data nodes,
+     but not for others, may result in the rejection of the
+     subscription.
+
+     The key words 'MUST', 'MUST NOT', 'REQUIRED', 'SHALL',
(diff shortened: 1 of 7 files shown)


src/netconf_subscribed_notifications.h
```diff
@@ -66,4 +66,13 @@ int np2srv_config_sub_ntf_filters_cb(sr_session_ctx_t *session, uint32_t sub_id,
...
         const char *path, const char *request_xpath, uint32_t request_id, struct lyd_node **parent, void *private_data);
 
+/**
+ * @brief Transform a datastore into a string identity.
+ *
+ * @param[in] str Identity.
+ * @param[out] ds Datastore.
+ * @return Sysrepo error value.
+ */
+const char *sub_ntf_ds2ident(sr_datastore_t ds);
```
michalvasko, line 76: Please move to `common.c/h`.

michalvasko (changes requested):
Okay, no problem, just do a minor refactoring and it should be good to go.

jktjkt:
> periodic-notifications-supported and on-change-supported: The publisher is capable of sending notifications for 'config false' and 'config true' nodes.
> minimum-update-period and minimum-dampening-period: The minimum value is 0.

When these fields are implemented, IMHO there's an expectation that the entire device (as visible over, say, NETCONF) will adhere to this behavior and send the updates based on the advertised intervals. But in case of sysrepo, I think this really depends on the way how your application-level CBs are set up -- isn't it? In that case, is it a good idea to possibly advertise features which are not met by the rest of the system?

jeremie6wind:
Hi,
thank you for your feedback.

I agree that the subscriptions set up defines the behavior of the notifications.
The idea here is to put default values for the fields.
For example, it is possible to configure 0 as the notification period for example, if nothing is implemented.
Other applications can subscribe to provide operational data and overwrite the values(define per node values)
according to the device capabilities thanks to the merge flag.
Would you prefer the content without the periods? without periodic-notifications-supported and on-change-supported?
Thanks

michalvasko (approved):

Issue #1741, opened 2025-06-04
mbedtls support for netopeer2.


amitdalodiya9:
Hi @michalvasko ,
We trying to upgrade sysrepo, libnetconf2, libyang and netopeer2 to latest release. Our platfoarm support only mbedtls (not supported openssl 3.x.x).
All module compilation is ok but netopeer2 seems still required openssl . 
Can you suggest how we can use mbedtls for netopeer2? if this is not possible current can we use openssl 1.x.x version to compile netopeer2 latest version (v 2.4.1). 
is there future plan to move netopeer2 also on mbedtls ?

Roytak:
Hi, can you try using the latest devel branch? You should be able to compile with mbedTLS only now.

amitdalodiya9:
@Roytak we have used released code(v2.4.1) :if we cherry-pick these changes(https://github.com/CESNET/netopeer2/pull/1744) over v2.4.1 will it works.

Roytak:
It should.

PR #1744, opened 2025-06-09
Add mbedTLS support

Roytak:
Fixes https://github.com/CESNET/netopeer2/issues/1741


amitdalodiya9:
@Roytak if we cherry-pick these changes https://github.com/CESNET/netopeer2/pull/1744 will it works.

PR #1790, opened 2026-03-11
Minor modifications to ietf-system-capabilities state RFC 9196

jeremie6wind:
Hi,

I propose those two patches to modify the content of the ietf-system-capabilities state.

For the first patch:
"/" xpath  is missing from the state.
As per-node-capabilities is a keyless list. Create the node-selector with lyd_new_term.
For the 2nd patch:
I removed the config-changes and state-changes, so another application can merge
custom xpaths with custom support for on-change and periodic notifications.

Thank you.
Best regards,
Jeremie

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -604,8 +604,15 @@ np2srv_capabilities_oper_cb(sr_session_ctx_t *session, uint32_t sub_id,
         }
 
         if (lyd_new_list(datastore_capas, NULL, "per-node-capabilities",
-                0, &per_node_capas, '/')) {
+                0, &per_node_capas, "")) {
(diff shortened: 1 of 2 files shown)


michalvasko:
I do not generally have a problem with the changes but since the subscriptions are handled by *sysrepo*, periodic subscriptions will always be supported for all the data. As for on-change, those may not be supported for some `operational` data but I am not sure it is really worth it to make it as complex as you suggest.

Regarding the failing tests, I was just about to fix it but okay, since you did.

jeremie6wind:
Ok so I will remove the patch in the middle.
Thank you

michalvasko (approved):

PR #1805, opened 2026-04-30
privcand UPDATE add private candidate datastore support

Ardival:
- include corresponding tests

### DIFF
--- a/modules/ietf-netconf-private-candidate@2026-02-03.yang
+++ b/modules/ietf-netconf-private-candidate@2026-02-03.yang
@@ -1,0 +1,111 @@
+module ietf-netconf-private-candidate {
+  yang-version 1.1;
+  namespace "urn:ietf:params:xml:ns:yang:ietf-netconf-private-candidate";
+  prefix pc;
+
+
+  organization
+    "IETF NETCONF (Network Configuration) Working Group";
+  contact
+    "WG Web:   <http://tools.ietf.org/wg/netconf/>
+     WG List:  <netconf@ietf.org>
+
+     Editor:   James Cumming
+            <james.cumming@nokia.com>
+
+     Editor:   Robert Wills
+            <rowills@cisco.com>";
+  description
+    "NETCONF private candidate support.
+
+     Copyright (c) 2026 IETF Trust and the persons identified as
+     authors of the code.  All rights reserved.
+
+     Redistribution and use in source and binary forms, with or
+     without modification, is permitted pursuant to, and subject to
+     the license terms contained in, the Revised BSD License set
+     forth in Section 4.c of the IETF Trust's Legal Provisions
+     Relating to IETF Documents
+     (https://trustee.ietf.org/license-info).
+
+     This version of this YANG module is part of RFC XXXX
+     (https://www.rfc-editor.org/info/rfcXXXX); see the RFC itself
+     for full legal notices.
+
+     The key words 'MUST', 'MUST NOT', 'REQUIRED', 'SHALL', 'SHALL
+     NOT', 'SHOULD', 'SHOULD NOT', 'RECOMMENDED', 'NOT RECOMMENDED',
+     'MAY', and 'OPTIONAL' in this document are to be interpreted as
+     described in BCP 14 (RFC 2119) (RFC 8174) when, and only when,
+     they appear in all capitals, as shown here.";
+
+  revision 2026-02-03 {
+    description
+      "Introduce private candidate support";
+    reference
+      "draft-ietf-netconf-privcand:
+       Netconf Private Candidates";
+  }
+  revision 2025-10-30 {
+    description
+      "Introduce private candidate support";
+    reference
+      "draft-ietf-netconf-privcand:
+       Netconf Private Candidates";
+  }
+  revision 2024-09-12 {
+    description
+      "Introduce private candidate support";
+    reference
+      "draft-ietf-netconf-privcand:
+       Netconf Private Candidates";
+  }
+
+  feature private-candidate {
+    description
+      "NETCONF :private-candidate capability;
+       If the server advertises the :private-candidate
+       capability for a session, then this feature must
+       also be enabled for that session.  Otherwise,
+       this feature must not be enabled.";
+    reference
+      "draft-ietf-netconf-privcand";
+  }
+
+  rpc update {
+    if-feature "private-candidate";
+    description
+      "Updates the private candidate from the running
+       configuration.";
+    reference
+      "draft-ietf-netconf-privcand";
+    input {
+      leaf resolution-mode {
+        type enumeration {
+          enum revert-on-conflict {
+            description
+              "Reject update when any conflicting
+               node is found and revert the private
+               candidate configuration datastore to its
+               state prior to issuing the update.";
+          }
+          enum prefer-candidate {
+            description
+              "Resolve conflicted node by selecting
+               the private candidate configuration
+               datastore version.";
+          }
+          enum prefer-running {
+            description
+              "Resolve conflicted node by selecting
+               the running configuration datastore
+               version.";
+          }
+        }
+        default "revert-on-conflict";
+        description
+          "Mode to resolve conflicts between running and
+           private-candidate configurations.";
+      }
+    }
+  }
+}
--- a/src/common.c
+++ b/src/common.c
@@ -740,6 +740,9 @@ np_new_session_cb(const char *UNUSED(client_name), struct nc_session *new_sessio
         goto error;
     }
 
+    user_sess->use_private_cand = nc_session_cpblt(new_session,
+            "urn:ietf:params:netconf:capability:private-candidate:1.0") ? 1 : 0;
+
     /* generate ietf-netconf-notification's netconf-session-start event for sysrepo */
     np_send_notif_session_start(new_session, np2srv.sr_sess, np2srv.sr_timeout);
 
@@ -1240,8 +1243,8 @@ cleanup:
 }
 
 struct nc_server_reply *
-np_op_filter_data_get(sr_session_ctx_t *session, uint32_t max_depth, uint32_t get_opts, const char *xp_filter,
-        struct lyd_node **data)
+np_op_filter_data_get(struct np_user_sess *user_sess, sr_datastore_t ds, uint32_t max_depth, uint32_t get_opts,
+        const char *xp_filter, struct lyd_node **data)
 {
     sr_data_t *sr_data = NULL, *sr_ln2_nc_server = NULL;
     struct lyd_node *e, *ignored_mod;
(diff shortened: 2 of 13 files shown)


modules/ietf-netconf-private-candidate@2026-02-03.yang
```diff
@@ -0,0 +1,111 @@
...
+          }
+        }
+        default "revert-on-conflict";
+        description
+          "Mode to resolve conflicts between running and
+           private-candidate configurations.";
+      }
+    }
+  }
+}
```
michalvasko: Missing newline.

src/common.c
```diff
@@ -41,6 +41,7 @@
 
 #include <libyang/libyang.h>
 #include <libyang/plugins_types.h>
+#include <nc_client.h>
```
michalvasko: Should not be needed.

src/common.c
```diff
@@ -1881,3 +1894,139 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+    }
+}
+
+struct nc_server_reply *
+np_reply_err_conflict(const struct lyd_node *rpc, sr_pc_conflict_set_t *conflict_set)
+{
+    struct lyd_node *err, *err_info_node;
+    char *run_val, *cand_val, *xpath = NULL;
+    uint32_t i;
+    int ret;
```
michalvasko: More accurate type would be `LY_ERR`. And temporary return values are usually called `r` to not be confused with function return values `ret` or more recently `rc`.

src/common.c
```diff
@@ -1881,3 +1894,139 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+
+    free(xpath);
+    sr_pc_free_conflicts(conflict_set);
+    return nc_server_reply_err(err);
+
+internal_error:
+    free(xpath);
+    sr_pc_free_conflicts(conflict_set);
+    lyd_free_tree(err_info_node);
+    return np_reply_err_op_failed(NULL, LYD_CTX(rpc), "Failed to build conflict error message.");
```
michalvasko: Any `err` may leak. Best to unify this into a single `cleanup` section that will check `ret` value.

src/common.h
```diff
@@ -57,6 +58,11 @@ struct np_user_sess {
     pthread_mutex_t lock;
 
     struct np_ntf_arg ntf_arg;
+
+    int use_private_cand;                   /* flag to use private candidate instead of shared candidate */
+    sr_priv_cand_t *private_ds;             /* private candidate structure */
+    sr_priv_cand_t *private_ds_backup;      /* backup of private candidate */
+    int privcand_lock;
```
michalvasko: Missing description.

src/common.h
```diff
@@ -450,4 +468,13 @@ struct nc_server_reply *np_reply_err_in_use(const struct ly_ctx *ly_ctx, const c
  */
 const char *sub_ntf_ds2ident(sr_datastore_t ds);
 
+/**
+ * @brief Build a NETCONF error message for conflicts during update.
```
michalvasko: Private candidate should be mentioned specifically.

src/netconf.c
```diff
@@ -210,6 +212,51 @@ np2srv_rpc_get_cb(const struct lyd_node *rpc, struct np_user_sess *user_sess)
...
+        reply = np_reply_err_sr(user_sess->sess, LYD_NAME(rpc));
+        goto cleanup;
+    }
+
+    if (sr_pc_edit_config(user_sess->sess, user_sess->private_ds, config, defop)) {
+        reply = np_reply_err_sr(user_sess->sess, LYD_NAME(rpc));
+        goto cleanup;
+    }
+
+    if (sr_pc_validate(user_sess->sess, NULL, user_sess->private_ds)) {
```
michalvasko: Candidate datastore is not required to be valid, private candidate should most likely work the same way.

src/common.h
```diff
@@ -450,4 +468,13 @@ struct nc_server_reply *np_reply_err_in_use(const struct ly_ctx *ly_ctx, const c
...
+ *
+ * @param[in] rpc Executed RPC.
+ * @param[in] conflict_set Set of conflicts.
+ * @return Server reply structure.
+ */
+struct nc_server_reply *np_reply_err_conflict(const struct lyd_node *rpc, sr_pc_conflict_set_t *conflict_set);
```
michalvasko: Seems `conflict_set` is spent. Not just a bad practice, it is not even documented.

tests/test_privcand.c
```diff
@@ -0,0 +1,948 @@
...
+    const char *data =
+            "<first xmlns=\"urn:ed1\" xmlns:xc=\"urn:ietf:params:xml:ns:netconf:base:1.0\" xc:operation=\"remove\"/>"
+            "<cont xmlns=\"urn:ed1\" xmlns:xc=\"urn:ietf:params:xml:ns:netconf:base:1.0\" xc:operation=\"remove\"/>";
+
+    SR_EDIT_SESSION(st, st->sr_sess, data);
+    FREE_TEST_VARS(st);
+    SR_EDIT_SESSION(st, st->sr_sess2, data);
+    FREE_TEST_VARS(st);
+
+    // temporary fix for <discard-config> by update and discard changes
```
michalvasko: ?

tests/test_privcand.c
```diff
@@ -0,0 +1,948 @@
...
+    assert_non_null(st->rpc);
+
+    st->msgtype = nc_send_rpc(st->nc_sess, st->rpc, 1000, &st->msgid);
+    assert_int_equal(st->msgtype, NC_MSG_RPC);
+
+    st->msgtype = nc_recv_reply(st->nc_sess, st->rpc, st->msgid, 3000, &st->envp, &st->op);
+    assert_int_equal(st->msgtype, NC_MSG_REPLY);
+    assert_non_null(st->envp);
+
+    lyd_print_mem(&st->str, st->envp, LYD_XML, 129);
```
michalvasko: Please use proper printing flags.

tests/test_privcand.c
```diff
@@ -0,0 +1,948 @@
...
+            "</get-config>\n";
+    assert_string_equal(st->str, expected);
+    FREE_TEST_VARS(st);
+
+    /* disconnect session, commit is rolled back */
+    sid = nc_session_get_id(ncs);
+    nc_session_free(ncs, NULL);
+
+    /* reply is sent before the server callback is called so give it a chance to perform the rollback */
+    usleep(100000);
```
michalvasko, line 782: Not enough to wait for the notification?

michalvasko (changes requested):

src/common.c
```diff
@@ -1881,3 +1893,138 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+            }
+
+            /* add conflict node into error msg*/
+            nc_err_add_info_other(err, err_info_node);
+        }
+    }
+
+cleanup:
+    free(xpath);
+    if (r) {
```
michalvasko, line 2024: May be uninitialized (if `conflist_set` is `NULL`).

src/common.c
```diff
@@ -1881,3 +1893,138 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+                        cand_val, NULL, "urn:ietf:params:xml:ns:yang:ietf-netconf-private-candidate", NULL);
+                free(cand_val);
+                if (r) {
+                    goto cleanup;
+                }
+
+            }
+
+            /* add conflict node into error msg*/
+            nc_err_add_info_other(err, err_info_node);
```
michalvasko, line 2016: Clearer and more readable code would be to move `err_info_node = NULL;` right after this line and also initializing `err_info_node` to `NULL`. Then `lyd_free_tree(err_info_node)` in the cleanup could be moved outside the `r` condition.

src/common.h
```diff
@@ -57,6 +58,11 @@ struct np_user_sess {
     pthread_mutex_t lock;
 
     struct np_ntf_arg ntf_arg;
+
+    int use_private_cand;                   /* flag to use private candidate instead of shared candidate */
+    sr_priv_cand_t *private_ds;             /* private candidate structure */
+    sr_priv_cand_t *private_ds_backup;      /* backup of private candidate */
+    int privcand_lock;                      /* lock for private candidate */
```
michalvasko: Please, mention the purpose at least a but more clearly. Something like `flag for tracking the lock state, changed by <lock> and <unlock> RPCs`.

src/netconf.c
michalvasko: Should be called only for `test-only`, just like in the standard `edit-config`.

src/common.h
michalvasko: `conflict_set` should now be `const`.

michalvasko (changes requested):

src/netconf.c
```diff
@@ -210,6 +212,52 @@ np2srv_rpc_get_cb(const struct lyd_node *rpc, struct np_user_sess *user_sess)
...
+        const struct lyd_node *config)
+{
+    struct nc_server_reply *reply = NULL;
+    sr_priv_cand_t *dup_privcand = NULL;
+
+    /* create private candidate if not yet created */
+    NP2_CHECK_PRIVCAND_EXISTS(user_sess, rpc, reply, cleanup);
+
+    /* duplicate private candidate for test-only */
+    if (sr_pc_backup_privcand(user_sess->sess, user_sess->private_ds, &dup_privcand)) {
```
michalvasko: Seems you need the backup only on `test-only` so no need to create it otherwise.

src/common.c
```diff
@@ -1881,3 +1893,138 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+    char *run_val = NULL, *cand_val = NULL, *xpath = NULL;
+    uint32_t i;
+    LY_ERR r = LY_SUCCESS;
+
+    err = nc_err(LYD_CTX(rpc), NC_ERR_OP_FAILED, NC_ERR_TYPE_APP);
+    nc_err_set_msg(err, "Update failed due to conflicts between private candidate and running configuration.", NULL);
+
+    if (conflict_set) {
+        for (i = 0; i < conflict_set->conflict_count; i++) {
+            err_info_node = NULL;
```
michalvasko: Strange to NULL the variable both at the loop beginning and at the end. This can be removed if the var is initialized to NULL.

michalvasko (changes requested):

src/common.c
```diff
@@ -1898,3 +1893,137 @@ sub_ntf_ds2ident(sr_datastore_t ds)
...
+    default:
+        val = lyd_get_value(node);
+        return val ? strdup(val) : NULL;
+    }
+}
+
+struct nc_server_reply *
+np_reply_err_conflict(const struct lyd_node *rpc, const sr_pc_conflict_set_t *conflict_set)
+{
+    struct lyd_node *err, *err_info_node;
```
michalvasko: As I wrote, you must initialize `err_info_node` to `NULL` to avoid invalid free in the cleanup.

michalvasko (changes requested):

michalvasko (approved):

PR #1806, opened 2026-05-11
common UPDATE allow custom error message

jeremie6wind:
Hello,
np_reply_err_sr checks the "(path" pattern in the error log to allocate the path variable.
When a RPC creates a custom error message without the pattern, the path variable is not allocated.
netopeer2-server can crash.
In this patch, don't use the path variable if not allocated and produce a default error on cleanup.

This is the case with some error messages from libyang-python.
BR


michalvasko (approved):

michalvasko:
Thanks, looks okay.

PR #1808, opened 2026-05-28
support multiple subscriptions to `/ietf-subscribed-notifications:streams`

jktjkt:
Currently, the NETCONF server makes an assumption that it will be the only entity which manages this container. With our RESTCONF server, this is no longer the case. Unless there's a systemwide "notification daemon" which will be in charge of this, the only reasonable choice is, IMHO, to rely on the usual rules of data merging as provided by sysrepo.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1306,7 +1306,12 @@ server_data_subscribe(void)
     }
 
     /* operational data */
-    SR_OPER_SUBSCR(mod_name, "/ietf-subscribed-notifications:streams", srsn_oper_data_streams_cb);
+    rc = sr_oper_get_subscribe(np2srv.sr_sess, mod_name, "/ietf-subscribed-notifications:streams",
+            srsn_oper_data_streams_cb, NULL, SR_SUBSCR_OPER_MERGE, &np2srv.sr_data_sub);
+    if (rc != SR_ERR_OK) {
+        ERR("Subscribing for providing \"%s\" state data failed (%s).", mod_name, sr_strerror(rc));
+        goto error;
+    }
     rc = sr_oper_get_subscribe(np2srv.sr_sess, mod_name, "/ietf-subscribed-notifications:subscriptions",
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
Hmm, okay, I suppose because I cannot think of a better solution. But this will result in `streams` with some streams that either *netopeer2* or *rousette* cannot actually subscribe to despite being advertised.

src/main.c
```diff
@@ -1320,7 +1306,12 @@ server_data_subscribe(void)
     }
 
     /* operational data */
-    SR_OPER_SUBSCR(mod_name, "/ietf-subscribed-notifications:streams", srsn_oper_data_streams_cb);
+    rc = sr_oper_get_subscribe(np2srv.sr_sess, mod_name, "/ietf-subscribed-notifications:streams",
+        srsn_oper_data_streams_cb, NULL, SR_SUBSCR_OPER_MERGE, &np2srv.sr_data_sub);
```
michalvasko: Double indent, 8 spaces, required.

michalvasko (changes requested):

src/main.c
jktjkt: done

jktjkt:
> Hmm, okay, I suppose because I cannot think of a better solution. But this will result in `streams` with some streams that either _netopeer2_ or _rousette_ cannot actually subscribe to despite being advertised.

If this becomes a problem, we have support for server-side filtering of returned data that's read from sysrepo. We are currently using that to provide [URLs to YANG schemas](https://github.com/CESNET/rousette/blob/master/src/restconf/YangSchemaLocations.cpp#L23) because these URLs depend on the hostname and server port which might be different in each incoming HTTP request. I can imagine that we might feed RESTCONF-specific URLs to only RESTCONF replies if it's ever needed.

michalvasko (approved):

michalvasko:
The only consequence for *netopeer2* right now will be returning an error if a user tries to subscribe to a *rousette* stream. Seems like a minor issue.

Issue #1810, opened 2026-06-02
`privcand` test is unstable


jktjkt:
This test tends to fail at times in our CI, especially when run with TSAN (which, in general, "makes the code run slower", and our CI runs all tests in parallel on a "big VM"). Here's one such failure ([full output](https://object-store.brno.openstack.cloud.e-infra.cz/swift/v1/KEY_b637b9c937414b29b3e277b4a85cc658/ci-logs-public/24/9224/3/check/deps-f38-clang-tsan/a7a789d/job-output.txt)):
```
16/19 Test #18: test_privcand ....................Subprocess aborted***Exception:   7.25 sec
[==========] tests: Running 13 test(s).
[ RUN      ] test_pc_copy_config
[       OK ] test_pc_copy_config
[ RUN      ] test_pc_update_rpc
[       OK ] test_pc_update_rpc
[ RUN      ] test_pc_validate
[       OK ] test_pc_validate
[ RUN      ] test_pc_nmda_rpcs
[       OK ] test_pc_nmda_rpcs
[ RUN      ] test_pc_lock_unlock
[       OK ] test_pc_lock_unlock
[ RUN      ] test_pc_basic_commit
[       OK ] test_pc_basic_commit
[ RUN      ] test_pc_commit_timeout_runout
Session 2 [ERR]: Communication file descriptor (10) unexpectedly closed.
0x8 != 0
[   LINE   ] --- /home/ci/src/cesnet-gerrit-public/CzechLight/dependencies/netopeer2/tests/test_privcand.c:461: error: Failure!
```
Our test runner is capturing the individual `np2.log` files from the test directory, [here's the relevant one](https://object-store.brno.openstack.cloud.e-infra.cz/swift/v1/KEY_b637b9c937414b29b3e277b4a85cc658/ci-logs-public/24/9224/3/check/deps-f38-clang-tsan/a7a789d/np2-test_privcand.log). I don't see any error in any of the logs, including [the syslog](https://object-store.brno.openstack.cloud.e-infra.cz/swift/v1/KEY_b637b9c937414b29b3e277b4a85cc658/ci-logs-public/24/9224/3/check/deps-f38-clang-tsan/a7a789d/journal.txt).

michalvasko:
I am used to server crashing on this error in the test and the output does seem truncated so I think that is what happened. Never seen it myself and I would need at least `valgrind` or `gdb` output.

PR #1811, opened 2026-06-03
tests: skip the `SIGEV_THREAD` code path in `test_privcand` under TSAN

jktjkt:
This particular check leads to a code path in netopeer2-server which starts a background thread via the C library, and that causes a segfault in TSAN's internals. Yay.

Fixes: https://github.com/CESNET/netopeer2/issues/1810
Bug: https://github.com/google/sanitizers/issues/1612

### DIFF
--- a/tests/test_privcand.c
+++ b/tests/test_privcand.c
@@ -31,6 +31,14 @@
 #include "np2_test.h"
 #include "np2_test_config.h"
 
+#if defined (__has_feature)
+#  if __has_feature(thread_sanitizer)
+#    define NP2_TSAN
+#  endif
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
Okay, I suppose, because it only concerns a test. But `gcc` requires also `defined(__has_feature)`, I think.

jktjkt:
> Okay, I suppose, because it only concerns a test.

Yeah, I opted for a smaller fix than disabling that test completely (which is already done for some other tests).

> But `gcc` requires also `defined(__has_feature)`, I think.

Done, I hope.

michalvasko (approved):

PR #1816, opened 2026-07-30
Add ietf-yp-transport-capabilities support

Roytak:
Implements draft-ietf-netconf-yp-transport-capabilities, which augments ietf-notification-capabilities to let clients discover supported notification transports, encodings (XML/JSON/CBOR), and security protocols (DTLS/TLS) at runtime instead of relying on error-based negotiation. Currently advertises UDP-Notif capabilities based on the enabled ietf-udp-notif-transport features.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -532,6 +532,114 @@ np2srv_content_id_cb(void *UNUSED(user_data))
 }
 
 /**
+ * @brief Add UDP-Notif transport capability to the transport-capabilities node.
+ *
+ * @param[in] transport_capas transport-capabilities container node.
+ * @param[in] yptc_mod ietf-yp-transport-capabilities module.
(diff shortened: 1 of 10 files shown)


src/main.c
```diff
@@ -645,6 +754,16 @@ np2srv_capabilities_oper_cb(sr_session_ctx_t *session, uint32_t sub_id,
         goto cleanup;
     }
 
+    /* transport capabilities */
+    if (!lyd_find_path(sys_capas, "ietf-notification-capabilities:subscription-capabilities", 0,
```
michalvasko: Redundant check.

michalvasko (changes requested):

michalvasko (approved):

PR #1823, opened 2026-09-07
server FEATURE hash a clear-text NETCONF user password before storing it

niklas-moser:
iana-crypt-hash, and RFC 9645 with it, define "$0$<cleartext>" as a value the server is expected to replace with a hash before storing. Netopeer2 stores it verbatim, leaving the clear text in the running datastore and in every export
of it. O-RAN WG11 R004, and 3GPP TS 33.117 clause 4.2.3.4.3.1 behind it, rule that out for a deployment that has to pass security conformance.

libnetconf2 can't fix this on its side: it gets the ietf-netconf-server data through a DONE_ONLY subscription, which sysrepo runs only after the commit. So this adds a second subscription with SR_SUBSCR_UPDATE — the one phase in
which a subscriber may still edit the data — and rewrites every created or modified "$0$" hashed-password under listen/ and call-home/ into a "$6$" digest, via nc_server_config_hash_password() from CESNET/libnetconf2#<N>.

Already hashed values are untouched, and so are older datastores still holding a "$0$" one, which the SSH auth path keeps accepting — no datastore migration, only a change to what gets stored from now on.

Depends on https://github.com/CESNET/libnetconf2/pull/639


src/main.c
```diff
@@ -1464,6 +1568,17 @@ server_data_subscribe(void)
...
+#ifdef NC_ENABLED_SSH_TLS
+    /* hash clear-text user passwords before they are stored, needs the "update" event which the
+     * DONE-only subscription below cannot get */
+    rc = sr_module_change_subscribe(np2srv.sr_sess, "ietf-netconf-server", NULL, np2srv_password_hash_update_cb,
+            NULL, 0, SR_SUBSCR_UPDATE, &np2srv.sr_data_sub);
+    if (rc != SR_ERR_OK) {
+        ERR("Subscribing for \"ietf-netconf-server\" password hashing failed (%s).", sr_strerror(rc));
+        goto error;
+    }
+#endif /* NC_ENABLED_SSH_TLS */
```
Roytak, line 1580: I think that this works only for the `running` datastore. You should probably add another subscription at least for the `startup` DS. Also should likely be `SR_SUBSCR_UPDATE | SR_SUBSCR_DONE_ONLY` if you don't care about the other events.

src/main.c
```diff
@@ -1227,6 +1227,110 @@ np2srv_libnetconf2_config_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id),
...
+            continue;
+        }
+
+        value = lyd_get_value(node);
+        if (!value || strncmp(value, "$0$", 3)) {
+            /* already a hash, nothing to do */
+            continue;
+        }
+
+        if (nc_server_config_hash_password(value + 3, &hashed_pw)) {
```
Roytak, line 1269: The value + 3 and possibly the check could be removed if you decide to integrate the comment in libnetconf2 PR.

Roytak (changes requested):

src/main.c
niklas-moser: Ok yes, I integrated it here https://github.com/CESNET/libnetconf2/pull/639

src/main.c
niklas-moser: I checked it and it also works for the `startup` datastore. Doesn't work yet for the `candidate` datastore. If you want I can also add that.

Added a second subscription on a `SR_DS_STARTUP` session and switched both to `SR_SUBSCR_UPDATE` | `SR_SUBSCR_DONE_ONLY`.

Roytak: Great, and that's right, `candidate` also needs it, so add that please. Otherwise the rest looks good.

niklas-moser: Good! I added support for the `candidate` datastore as well.

Roytak (approved):

michalvasko (approved):

PR #1829, opened 2026-09-18
Adding parts of the code that will work with openssl4

mruprich:
Adding similar changes to what can be found in libnetconf2 devel branch, only conditioned for more general usage with different openssl versions. Let me know if this looks good or not.


michalvasko (changes requested):
Are those `#if`s required? At least OpenSSL 3.0 is required, which should support the functions you added. Also, please change the target branch to `devel`, we do not accept PRs into `master`.

mruprich:
TBH, I did not try without those #ifs, let me try that and update the PR.

mruprich:
You're right, this works without the conditional openssl versioning.

michalvasko (approved):
Thanks, now it looks okay.

mruprich:
Thanks!

